PR #2, opened 2021-09-30
Removed parking-lot and used tokio::sync instead

Lesiuk:
[work in progress]

The goal of this PR is to remove all of the blocking operations and replace them with non-blocking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "15af2628f6890fe2609a3b91bef4c83450512802e59489f9c1cb1fa5df064a61"
 
 [[package]]
+name = "async-trait"
+version = "0.1.51"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "44318e776df68115a881de9a8fd1b9e53368d7a4a5ce4cc48517da3393233a5e"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -239,7 +250,6 @@ dependencies = [
  "indexmap",
  "librqbit_core",
  "log",
- "parking_lot",
  "pretty_env_logger",
  "rand 0.8.4",
  "serde",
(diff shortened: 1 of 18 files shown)


ikatson:
Wow, that's impressive work, thanks! I'm on phone now, can take a closer look at the change itself later. Can you provide more rationale for this, in what contexts is it better, and how? Did you measure any performance improvements maybe in some usecases?

Lesiuk:
Asynchronous locking primitives (like std Mutex or parking_lot Mutex) are somewhat incompatible with Rust's async model. Locking such Mutex or RwLock will block the executor thread from doing any other work. You should always use tokio::sync locking primitives inside tokio executor. Switching from parking_lot to tokio::sync will enable rqbit to max out multicore cpu in the future.

ikatson:
I don't think it's as clear-cut, as short-lived locks should be alright, and all or most of them here are. My main concern using async ones is that it might end up slower in the end (as for short-lived locks with low contention async ones are doing more work), and if so I wonder how much.

If it doesn't affect performance much though, or even makes it better, I'm all in to merge it. In addition, as you say, semantically it's more correct to use all-async approach. Need to take a look and measure some things.

Lesiuk:
They are a little slower, but this performance probably doesn't matter, and I have always been taught that you should never block inside the executor thread. They could be short lived but tokio executor have only 2 * core count threads by default. If there are few of them held at the same time, It limits your maximum throughput.

ikatson:
Ok, I looked at the code and it's beautiful. Thanks a lot for working on this.

Only one concern - async_trait usage (for PeerConnectionHandler) in the hot path. That implies frequent boxing as far as I understand (for Pin<Box<dyn Future>>>).

I don't see an easy way around it though, any ideas? I don't think it's blocking the merge though, we can figure smth out later if it becomes a bottleneck. 

In the meantime trying to remember how did I profile this a few months ago, to compare the difference before and after this change.

ikatson:
There's some kind of a deadlock or whatever happening though, takes 100% CPU. Noticed that when profiling. We need to fix this one before the merge, as rqbit doesn't work properly with this bug.

Can reproduce (after a couple tries) from the command line, here's how I tested

```
./target/release/rqbit ./crates/librqbit/resources/ubuntu-21.04-live-server-amd64.iso.torrent /tmp/torrent-download/ --overwrite -s
```
Screenshot 2021-10-01 at 10 21 02

ikatson:
Fully unwrapped profiler tree in case it helps. Seems like it tries to acquire some semaphore forever and wastes CPU on this.

Screenshot 2021-10-01 at 10 26 17

ikatson:
Looks like that's all due to the use of block_on within poll_next in PeerStream

ikatson:
I hacked around the bug in this branch  https://github.com/ikatson/rqbit/tree/Lesiuk-async-locks

And timed the results [in this spreadsheet](https://docs.google.com/spreadsheets/d/1_V88VJ99zsGd-jOxL_aoaCvBpHoUlV3LjEU6-M_sT3s/edit?usp=sharing)

The new version takes about 40-50% more CPU, which isn't that great. The average actual runtime to download a torrent obviously fluctuates due to networking, but it is in the same ballpark.

Lesiuk:
That's more than I thought It's going to be. I am going to investigate around this issue this weekend.

ikatson:
Closing for staleness

PR #29, opened 2023-08-09
clap replace deprecated attributes

imp:
clap v4 series deprecated some of the clap v3 derive attributes, while still supporting them for the sake of backward compatibility. In order to help migration it provides feature `deprecated`, which actually emits deprecation warning when these old attrs are used.

These PR enables that feature and also updates all the occurrences of the deprecated attributes with the current ones.

Having the `deprecated` feature enabled may be useful going forward to spot these things early.


ikatson:
Thanks for the cleanups!

PR #33, opened 2023-11-15
Code example for librqbit

ALEZ-DEV:
I have edited the readme of the librqbit library to showcase the library
This is a really simple program, but it work

### DIFF
--- a/crates/librqbit/README.md
+++ b/crates/librqbit/README.md
@@ -1,1 +1,128 @@
-../README.md
+# librqbit
+
+A torrent library 100% written in rust
+
+## Basic example
+This is a simple on how to use this library  
+This program will just download a simple torrent file with a Magnet link
+
+```rust
+use std::error::Error;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use librqbit::{http_api_client, Magnet};
+use librqbit::session::{AddTorrentResponse, ListOnlyResponse, ManagedTorrentState, Session};
+use librqbit::spawn_utils::BlockingSpawner;
+use size_format::SizeFormatterBinary as SF;
+use tokio::spawn;
+
+const MAGNET_LINK: &str = "magnet:?..."; // Put your magnet link here
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>>{
+
+    let spawner = BlockingSpawner::new(false);
+
+    // This function will print the torrent properties every 1s
+    let stats_printer = |session: Arc<Session>| async move {
+        loop {
+            session.with_torrents(|torrents| {
+                for (idx, torrent) in torrents.iter().enumerate() {
+                    match &torrent.state {
+                        ManagedTorrentState::Initializing => {
+                            println!("[{}] initializing", idx);
+                        },
+                        ManagedTorrentState::Running(handle) => {
+                            let peer_stats = handle.torrent_state().peer_stats_snapshot();
+                            let stats = handle.torrent_state().stats_snapshot();
+                            let speed = handle.speed_estimator();
+                            let total = stats.total_bytes;
+                            let progress = stats.total_bytes - stats.remaining_bytes;
+                            let downloaded_pct = if stats.remaining_bytes == 0 {
+                                100f64
+                            } else {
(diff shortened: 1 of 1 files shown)


crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+## Basic example
+This is a simple on how to use this library  
+This program will just download a simple torrent file with a Magnet link
+
+```rust
+use std::error::Error;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use librqbit::{http_api_client, Magnet};
```
ikatson, line 14: http_api_client isn't needed here

crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+use tokio::spawn;
+
+const MAGNET_LINK: &str = "magnet:?..."; // Put your magnet link here
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>>{
+
+    let spawner = BlockingSpawner::new(false);
+
+    // This function will print the torrent properties every 1s
```
ikatson, line 27: this block contributes to example size a lot, but isn't required for the example to work (to download the torrent).
I'd remove it to showcase the simplicity of use

crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+                    }
+                }
+            });
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    };
+
+    // Create the torrent session, one session can have more than one torrent at the same time
+    let session = Arc::new(
+        Session::new(PathBuf::from("C:\\Anime"), spawner).await?
```
ikatson, line 72: Maybe we can modify Session::new to take ```impl AsRef<Path>``` as argument, or at least use ```"C:\\Anime".into()``` for brevity

crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+    let handle = match session.add_torrent(MAGNET_LINK, None).await {
+        Ok(v) => match v {
+            AddTorrentResponse::AlreadyManaged(handle) => {
+                println!(
+                    "torrent {:?} is already managed, downloaded to {:?}",
+                    handle.info_hash, handle.output_folder
+                );
+                Err(())
+            }
+            AddTorrentResponse::ListOnly(ListOnlyResponse {
```
ikatson, line 89: there's no way to pass "listOnly" in the example code, so this branch can be removed

crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+    );
+
+    // Spawn the properties printer function
+    spawn(stats_printer(session.clone()));
+
+    // Add the magnet link to the torrent session, you don't have to specify if it's an url or magnet,
+    // the library will recognize automatically for you
+    let handle = match session.add_torrent(MAGNET_LINK, None).await {
+        Ok(v) => match v {
+            AddTorrentResponse::AlreadyManaged(handle) => {
```
ikatson, line 82: with a brand new session (as in this example), this shouldn't happen, so we can remove it

crates/librqbit/README.md
```diff
@@ -1 +1,128 @@
...
+            }
+        }
+        Err(err) => {
+            eprintln!("error adding {}: {:?}", MAGNET_LINK, err);
+            Err(())
+        }
+    };
+
+    // Wait until the session complete the torrent download and terminate
+    let _ = match handle {
```
ikatson, line 121: seems you can just simplity the rest of the example to

```handle?.wait_until_completed().await?```

**Review by ikatson — CHANGES_REQUESTED** [review]
Considering all the other comments, and with a couple code modifications, I think we can simplify the example to just


```
let session = Session::new("C:\\Anime", Default::default());
let handle = match session.add_torrent(MAGNET_LINK, None).await {
    AddTorrentResponse::Added(handle) => {
        handle
    },
    resp => unimplemented!("{:?}", resp)
}
handle.wait_until_completed().await?;

ALEZ-DEV:
There is the simplified version

ikatson:
Didn't check if it compiles, assume you did.
LGTM and thanks for the contribution!

PR #55, opened 2023-12-13
Chunk-piece length comparison fix

baltarifcan:
Fix for the case of piece length being equal to chunk length

Before:
Screenshot 2023-12-13 184621

After:
Screenshot 2023-12-13 185654


ikatson:
Thank you!

PR #56, opened 2023-12-13
Torrent status icon

arccik:
- Added a new component StatusIcon to handle the display of icons based on the state of the torrent.
- Modified the logic to dynamically choose the appropriate icon.
- Updated the color of the icons to enhance visual representation.


crates/librqbit/webui/src/components/StatusIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+} from "react-icons/md";
+
+type Props = {
+  finished: boolean;
+  isDownloading: boolean;
+  error: boolean;
+};
+
+export const StatusIcon = ({ finished, isDownloading, error }: Props) => {
+  const isSeeding = finished && isDownloading;
```
ikatson, line 17: you can't be both finished and downloading. You can be "finished" and "live" in backend terms

ikatson (changes requested):
Would be cool if you attach screenshots next time :)

Issue #70, opened 2023-12-24
Support BitTorrent v2 (BEP 52)


jabedude:
Love this project! I think it's one of the best rust BT implementations :) Just wanted to see if there's interest in supporting [BEP 52](http://bittorrent.org/beps/bep_0052.html). The major improvement is using a larger hash for security from hash collisions

ikatson:
Thanks for kind words @jabedude!

As for BTv2, in a nutshell - I'm not against it, but I'm not convinced it's where time should be put in next.

I looked through the BTv2 spec and it seems quite a change to implement it. The current architecture would need some heavy refactoring to support both v1 and v2.

I don't even know if I've ever seen V2-only torrents in the wild. Intuitively, when a significant % of torrents switches to V2 only, then (or a bit earlier) would be the time to add support for it, as otherwise things won't work. Maybe I don't know enough about BTv2, but I draw a parallel with HTTP 1 -> HTTP 2/3 - yes, it may be better, but it's singificantly different, harder to implement, and, most importantly, HTTP/1 was and will be supported many years since 2/3 appeared.

As for security against hash collisions being the major reason to do so, for me personally, it's not very compelling to justify putting time into it.

However, all that said, looks like you have some interest in actually implementing it (as you started in #72 ), so I'm ready to support you along the way. If the end result doesn't make things terribly complicated, I'd be happy to merge.

angrynode:
> As for security against hash collisions being the major reason to do so

That's not a bad reason, but that's definitely not the major reason in my opinion. Bittorrent v2 is great (on paper) because pieces are file-delimited (« per-file hash trees ») which enables cross-seeding of the same files/pieces across torrents.

There is strong untapped potential for better Bittorrent v2 integration where the same peer swarm could unknowingly serve "old" files part of new torrents (note i'm not talking about BEP44/BEP46 mutable torrents). For example, for archival processes and scientific datasets where new files are appended regularly but the old files don't ever change, and where currently Bittorrent is not a great fit because each new torrent revision looses the old seeds.

Example usecase (inspired by real-life events):

- a hypothetical non-profit RustTorrent foundation organizes a conference every year, and recordings from the talks are shared via Bittorrent v2:
- there's one torrent per year, per quality-setting (eg. 2025-480p, 2025-1080p, 2024-480p, 2024-480p, etc)
- there's also one torrent per year gathering all videos (eg. 2025-yearly), and also one torrent with the whole archive up to a year (2025-archive)
[...]

PR #72, opened 2023-12-27
Add support for BTv2 magnet links

jabedude:
This is a tiny step to supporting BEP52 (#70). Ideally the new `InfoHash` type would replace `Id20` in `Magnet`, but that would be a significant refactor across the project and it's not required until we actually begin performing lookups with SHA-256 hashes


ikatson:
Hey thanks for this, I'm traveling with almost no internet and only a phone, so will take a look in a few days when I come back.

crates/librqbit_core/src/magnet.rs
```diff
@@ -54,11 +77,90 @@ impl std::fmt::Display for Magnet {
     }
 }
 
+/// A parsed magnet link, supporting both v1 and v2 links.
+pub struct MagnetBEP52 {
```
ikatson, line 81: i'd probably go a bit further, and change the existing "Magnet" struct to contain "InfoHash" instead of "Id20". Whatever expects only Id20 would say "I only support BTV1 info hashes" or smth like that.

From memory (not looking at the code), the only place that works with magnet links is session, so you can just stick the match in there, or even better, make a ```fn as_id20(&self) -> Option<Id20>``` method on InfoHash, and then in e.g session add ```.context("magnet link doesn't contain BTv1 info hash")```

ikatson (changes requested):
Hey, first pass mostly LGTM, see the comment.
Without that suggestion implemented there'll just be a dead struct in the code

crates/librqbit_core/src/hash_id.rs
```diff
@@ -142,8 +142,42 @@ impl PartialOrd<Id20> for Id20 {
     }
 }
 
+/// A 32-byte hash used in Bittorrent V2, for torrent info hashes, piece hashing, etc.
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
+pub struct Id32(pub [u8; 32]);
```
ikatson, line 147: nit: can both of them be implemented with a const generic for the length? Would reduce the source code size as Id20 and Id32 are a copy-paste of each other otherwise.

crates/librqbit_core/src/magnet.rs
```diff
@@ -2,11 +2,49 @@ use std::str::FromStr;
 
 use anyhow::Context;
 
-use crate::id20::Id20;
+use crate::hash_id::{Id20, Id32};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum InfoHash {
+    V1(Id20),
```
ikatson, line 9: I agree that a single info-hash can be an enum of either V1 or V2.
However, if you imagine a magnet link, can a magnet link exist which provides both hashes? If so, the current Magnet struct doesn't account for that. Again, if that is true, maybe change Magnet to be ```struct Magnet {id20: Option<Id20>, id32: Option<Id32>}``` and move the "as_id20" like method into Magnet rather than InfoHash

crates/librqbit_core/src/magnet.rs
```diff
@@ -17,15 +55,20 @@ impl Magnet {
...
         let mut trackers = Vec::<String>::new();
         for (key, value) in url.query_pairs() {
             match key.as_ref() {
-                "xt" => match value.as_ref().strip_prefix("urn:btih:") {
-                    Some(infohash) => {
-                        info_hash.replace(Id20::from_str(infohash)?);
+                "xt" => {
+                    if let Some(ih) = value.as_ref().strip_prefix("urn:btih:") {
+                        let i: InfoHash = InfoHash::V1(Id20::from_str(ih)?);
+                        info_hash.replace(i);
```
ikatson, line 65: this "replace" doesn't account for both hashes being possible in the URL. I guess there's nothing that prevents both v1 and v2 hashes being possible in the URL. I don't know if there's a spec about it, but intuitively it makes total sense to have both in the same URL.

ikatson:
@jabedude left a few pretty significant design comments

crates/librqbit_core/src/magnet.rs
jabedude: >  can a magnet link exist which provides both hashes?

Very much yes, that was a silly oversight on my part. They're called hybrid magnet links, they look like this:
```
magnet:?xt=urn:btih:631a31dd0a46257d5078c0dee4e66e26f73e42ac&xt=urn:btmh:1220d8dd32ac93357c368556af3ac1d95c9d76bd0dff6fa9833ecdac3d53134efabb&dn=bittorrent-v1-v2-hybrid-test
```

The only issue with `struct Magnet {id20: Option<Id20>, id32: Option<Id32>}` is that it would be possible to represent a magnet link without an infohash if both are `None`. Not a huge deal but something to keep in mind

ikatson: That's fine, as long as magnet construction / parsing fails when there's 0 hashes, and if the fields are private, it would be impossible to construct a broken magnet.

ikatson:
Code LGTM so far, but:
- I assume you want to change Magnet too, so I'll wait for it before merging
- A bunch of tests got broken by this PR, PTAL

jabedude:
Tests fixed, changed `Magnet` to look how we discussed

crates/librqbit_core/src/hash_id.rs
```diff
@@ -37,94 +86,67 @@ impl Serialize for Id20 {
...
-        impl<'de> serde::de::Visitor<'de> for Visitor {
-            type Value = Id20;
+        struct IdVisitor<const N: usize>;
+
+        impl<'de, const N: usize> serde::de::Visitor<'de> for IdVisitor<N> {
+            type Value = Id<N>;
 
             fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
-                write!(formatter, "a 20 byte slice or a 40 byte string")
+                formatter.write_str("a byte array of length ")
```
ikatson, line 100: nit: could have just done ```formatter.write_fmt(format_args!("a byte array of length {}", N)))```

crates/librqbit_core/src/magnet.rs
```diff
@@ -2,41 +2,61 @@ use std::str::FromStr;
...
-                    None => anyhow::bail!("expected xt to start with urn:btih:"),
                 },
                 "tr" => trackers.push(value.into()),
                 _ => {}
             }
         }
-        match info_hash {
-            Some(info_hash) => Ok(Magnet {
-                info_hash,
+        match info_hash_found {
```
ikatson, line 53: nit: you don't need a separate variable to track this, could have just check if at least one of id20 and id32 is set

crates/librqbit_core/src/magnet.rs
```diff
@@ -45,20 +65,52 @@ impl Magnet {
 
 impl std::fmt::Display for Magnet {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "magnet:?xt=urn:btih:{}&tr={}",
-            self.info_hash.as_string(),
-            self.trackers.join("&tr=")
-        )
+        if let (Some(id20), Some(id32)) = (self.id20, self.id32) {
```
ikatson, line 68: nit: it's a bit repetitive to handle all cases like this, you could have done this without a big branch, but small branches instead:

```
write!(f, "magnet:?")?;
if let Some(id20) = self.id20 {
    write!("&...");
}
if let Some(id32) = self.i32 {
    write!("&...");
}
write!("&tr={}", self.trackers.join("&tr="))?;
```

crates/librqbit_core/src/magnet.rs
```diff
@@ -45,20 +65,52 @@ impl Magnet {
...
+                self.trackers.join("&tr=")
+            )
+        } else if let Some(id32) = self.id32 {
+            write!(
+                f,
+                "magnet:?xt=urn:btmh:1220{}&tr={}",
+                id32.as_string(),
+                self.trackers.join("&tr=")
+            )
+        } else {
```
ikatson, line 90: please don't panic in Display trait. If you restructure the code like I suggested above, let it just print a magnet without hashes

ikatson (approved):

ikatson:
All comments are minor, we can fix them later, so merging

PR #84, opened 2024-02-18
Add Dockerfile and Compose

marcosrosse:
Hello 👋🏿

This PR is to add a Dockerfile to use with rqbit. I'm currently using it in a Raspberry pi 4 and in laptop.

Things that need to improve is:

- Change the DTH and the session path for persistence.
- Change the Docker registry repository used for this image.
- Build an image for each OS architecture.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM curlimages/curl:latest AS downloader
+
+RUN ARCH=$(uname -m) \
+    && curl -L \
+        https://github.com/ikatson/rqbit/releases/latest/download/rqbit-linux-static-$ARCH > /tmp/rqbit \
+    && chmod +x /tmp/rqbit
+
+FROM debian:bullseye-slim AS final
+
+ARG UID=1000
+ARG GID=1000
+
+RUN groupadd -g "${GID}" appuser \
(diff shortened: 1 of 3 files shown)


ikatson:
Hi @marcosrosse, thanks for your contribution.

There are a few things that make me uncomfortable merging this.

## 1. CI should be building this
1. I can't reference a docker image "docker.io/marcosrossem/rqbit:latest", as it's not under my control.
2. I don't have docker and thus can't support this going forward.

However I would be fine merging it if the PR would make Github Actions build and publish the docker image automatically to docker.io.
For you to test this, I imagine you'd need to fork the repo, set up a Github Action that publishes to docker.io

## 2. Hardcoded command isn't flexible
It would be better if the entrypoint was "rqbit", but user could override the command as it's now hardcoded to do smth specific (e.g. start a server).

That's it for the first review round.
Alternatively, you can create a repo like rqbit-docker or smth like that and publish the docker support there.

infowolfe:
@ikatson @marcosrosse the default with github actions is to use ghcr.io (github packages), which would allow control to stay with the repo.

ikatson:
Added docker builds to main CI. Not yet multiarch, but will do that too later

PR #87, opened 2024-02-27
Support private torrents

av-gal:


ikatson:
This looks good so far, but I thought the trickier and necessary parts where:
- connect to only one tracker at once
- if the tracker is unresponsive, disconnect all peers before switching the tracker

Also, after a recent change there's a Todo to actually report to trackers the current torrent status (it's stubbed at the moment to report zeros).

I also don't have any private torrents / trackers to test this with.

I don't mind merging this as is, as it seems better than nothing for handling private.

@av-gal what do you think else is needed to mark private torrents supported?

ikatson:
Oh I didn't notice it's a draft PR, but the above question still applies

av-gal:
I wouldn't merge private tracker support until ratio reporting is fully functional. Private trackers often monitor each peer's upload and download amounts for their own purposes and might ban peers who they believe are trying to cheat their systems.

I think the main issue with peers for now is that rqbit doesn't have a notion of trackers "failing". A future is created by tracker_comms to monitor each tracker, but those futures only return if there's an error in creating the request. A tracker failing to connect or returning an error code only causes that future to increase the time it sleeps in between each request. 

I think I might try to refactor it by creating one PeerSource trait for each method of peer discovery. This would  allow us to reuse connection information across many torrents that may be connected to one tracker, and moving the retry behaviour to operate on that trait would also allow the caller to decide retry behaviour based on if the torrent is private or not. 

Do you have any thoughts on that design? I might try to do that refactor this week.

ikatson:
A couple things

1. On your redesign proposal, I'm afraid I didn't fully understand the problem (the way you put it) or the idea of how to approach it. But at the same time as long as it helps solve private trackers and isn't too complicated I don't mind it.
2. I'm travelling this week and have very limited availability. So if you plan to do it this week, I won't be able to provide quick feedback, so use your judgement.

Thanks!

ikatson:
@av-gal I tried to draft private torrent support on the airplane today. It's not fully finished, but almost. https://github.com/ikatson/rqbit/compare/main...private

I'm not a big fan of how it paned out, but the core ideas are the following:
1. tracker comms works a bit differently in private mode. It loops only one tracker at a time, and when it fails, it calls forget_all_peers() on the torrent (through TorrentActionsProvider trait to keep tracker_comms isolated).
2. forget_all_peers() didn't seem easy, so instead I made it pause, then unpause.
3. However pausing destroys the PeerStream, so I shoved a kludge there to return the peer stream instead of destroying it on pause(). And then unpause() accepts an optional peer stream.

Let me know what you think and if your idea was cleaner.

infowolfe:
@ikatson would it be possible to rebase the private tracker support branch? it's super out of date.

PR #96, opened 2024-03-22
README: Add repology packaging status

kpcyrd:


ikatson:
Oh wow I didn't know someone packaged it for Arch, cool!
Thanks for this

PR #97, opened 2024-03-24
Make the announce info field optional

anacrolix:
BEP 3 makes it seem like announce isn't optional, but in practice it is. This relaxes the constraint and makes a whole bunch of torrents that it borked on work.

I wasn't sure about the fallback with http://retracker.local/announce, I'm guessing that's not necessary if announce can now be optional.

I noticed Either is imported, but referred to in that part of the code with its full path. I left it alone, I think it's more readable knowing it's from itertools.


ikatson:
Looks good, thanks!

PR #99, opened 2024-03-26
Remove info hash hacks and misc

anacrolix:
I ran into this same problem in my client https://github.com/anacrolix/torrent, there I solved by implementing the equivalent of a deserializer into bytes (essentially just type checks the bencode) (https://github.com/anacrolix/torrent/blob/master/metainfo/metainfo.go#L16). I don't quite know enough Rust, but I think you either want to (in the same way as my implementation) store the info bytes directly in a metainfo so you can hash them at will (and do a subsequent deserialize if you want to turn them into an info you can work with), or as I've done here, I've stored info as a BencodeValue, which can be deterministically turned into bytes for hashing, or deserialized further into an Info. I did notice you use HashMap for BencodeValue::Dict. That shouldn't allow serializing back into a different order, but I couldn't trigger that behaviour with the unit test. I'm not sure how to best trigger this behaviour in Rust. The solution would be to have an ordered mapping type, or have Dict use Vec internally and expose some kind of mapping interface? BitTorrent does specify that dict keys must be ordered like memcmp (but not in the original BEP 3), so client support is flakey, and that's why you want to deserialize info into something that retains the original ordering, at least if you intend to generate infohashes.

Moving the infohashing out of bencode moved the sha1w dep to librqbit_core. I didn't retain the features, I assume they aren't in use (except on the main bin perhaps?), but if it was I can put them back in mapping features to core instead.

There's no way to deserialize from BencodeValue to a Rust type. I'm not well versed in serde, would that require a custom deserializer? Instead I serialized back into bytes then ran the bytes deserializer. It should still be correct, unless the Dict nondeterministic ordering strikes. At least for infohashing, there's no intermediate allocation as it can serialize straight into the hasher.

How does one deserialize into bytes with serde? If it was possible to have `info: ByteBuf` in the metainfo that could be optimal for hashing after deserializing. With the BencodeValue route I went, I had to use BencodeValueOwned: I couldn't work out the trait and lifetime requirements to have it be BencodeValue<ByteBuf>. Do you know why that might be? Perhaps knowing whether BencodeValue or a raw ByteBuf type could be used in metainfo to avoid allocations could decide the best choice?

I tried to add a std::io::Write for any ISha1, but Rust didn't like that, so I made a helper type then noticed you had a private one in a function somewhere that was identical.

I have been toying with the idea of a Rust implementation of my torrent client for a while, but there's a few rough spots in my knowledge of Rust. If it works for you'll I'll make some more contributions like this if you have the capacity for it.

### DIFF
--- a/crates/bencode/Cargo.toml
+++ b/crates/bencode/Cargo.toml
@@ -15,6 +15,7 @@ serde = {version = "1", features=["derive"]}
 buffers = {path = "../buffers", package="librqbit-buffers", version = "2.2.1"}
 clone_to_owned = {path = "../clone_to_owned", package="librqbit-clone-to-owned", version = "2.2.1"}
 anyhow = "1"
+serde_bytes = "0.11.14"
 
 [dev-dependencies]
 serde_json = "1.0.115"
--- a/crates/bencode/Cargo.toml
+++ b/crates/bencode/Cargo.toml
@@ -15,3 +15,6 @@ serde = {version = "1", features=["derive"]}
 buffers = {path = "../buffers", package="librqbit-buffers", version = "2.2.1"}
 clone_to_owned = {path = "../clone_to_owned", package="librqbit-clone-to-owned", version = "2.2.1"}
 anyhow = "1"
+
+[dev-dependencies]
+serde_json = "1.0.115"
--- a/crates/bencode/Cargo.toml
+++ b/crates/bencode/Cargo.toml
@@ -10,15 +10,8 @@ readme = "README.md"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[features]
-default = ["sha1-system"]
-sha1-system = ["sha1w/sha1-system"]
-sha1-openssl = ["sha1w/sha1-openssl"]
-sha1-rust = ["sha1w/sha1-rust"]
-
 [dependencies]
 serde = {version = "1", features=["derive"]}
 buffers = {path = "../buffers", package="librqbit-buffers", version = "2.2.1"}
 clone_to_owned = {path = "../clone_to_owned", package="librqbit-clone-to-owned", version = "2.2.1"}
 anyhow = "1"
-sha1w = {path="../sha1w", default-features=false, package="librqbit-sha1-wrapper", version="2.2.1"}
--- a/crates/bencode/src/bencode_value.rs
+++ b/crates/bencode/src/bencode_value.rs
@@ -96,7 +96,7 @@ where
 // A dynamic value when we don't know exactly what we are deserializing.
 // Useful for debugging.
 
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Clone)]
 pub enum BencodeValue<BufT: std::hash::Hash + Eq> {
     Bytes(BufT),
     Integer(i64),
--- a/crates/bencode/src/bencode_value.rs
+++ b/crates/bencode/src/bencode_value.rs
@@ -1,19 +1,19 @@
 use std::{collections::HashMap, marker::PhantomData};
 
-use buffers::{ByteBuf, ByteString};
+use buffers::{ByteBuf, ByteBufT, ByteString};
 use clone_to_owned::CloneToOwned;
 use serde::Deserializer;
 
-use crate::serde_bencode_de::from_bytes;
+use super::*;
 
-pub fn dyn_from_bytes<'de, BufT>(buf: &'de [u8]) -> anyhow::Result<BencodeValue<BufT>>
+pub fn dyn_from_bytes<'a, BufT>(buf: &'a [u8]) -> anyhow::Result<BencodeValue<BufT>>
 where
-    BufT: From<&'de [u8]> + std::hash::Hash + Eq,
+    BufT: BencodeValueBufConstraint + From<&'a [u8]>,
 {
     from_bytes(buf)
 }
 
-impl<BufT: serde::Serialize + Eq + std::hash::Hash> serde::Serialize for BencodeValue<BufT> {
+impl<BufT: serde::Serialize + BencodeValueBufConstraint> serde::Serialize for BencodeValue<BufT> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
@@ -29,7 +29,7 @@ impl<BufT: serde::Serialize + Eq + std::hash::Hash> serde::Serialize for Bencode
 
 impl<'de, BufT> serde::de::Deserialize<'de> for BencodeValue<BufT>
 where
-    BufT: From<&'de [u8]> + std::hash::Hash + Eq,
+    BufT: BencodeValueBufConstraint + From<&'de [u8]>,
 {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
@@ -41,7 +41,7 @@ where
 
         impl<'de, BufT> serde::de::Visitor<'de> for Visitor<BufT>
         where
-            BufT: From<&'de [u8]> + std::hash::Hash + Eq,
+            BufT: BencodeValueBufConstraint + From<&'de [u8]>,
         {
             type Value = BencodeValue<BufT>;
 
@@ -56,6 +56,13 @@ where
                 Ok(BencodeValue::Integer(v))
             }
 
+            fn visit_borrowed_bytes<E>(self, v: &'de [u8]) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                Ok(BencodeValue::Bytes(BufT::from(v)))
+            }
+
             fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
             where
                 A: serde::de::SeqAccess<'de>,
@@ -65,13 +72,6 @@ where
                     v.push(value);
                 }
                 Ok(BencodeValue::List(v))
-            }
-
-            fn visit_borrowed_bytes<E>(self, v: &'de [u8]) -> Result<Self::Value, E>
-            where
-                E: serde::de::Error,
-            {
-                Ok(BencodeValue::Bytes(BufT::from(v)))
             }
 
             fn visit_map<A>(self, mut map: A) -> Result<Self::Value, A::Error>
@@ -93,18 +93,32 @@ where
     }
 }
 
-// A dynamic value when we don't know exactly what we are deserializing.
-// Useful for debugging.
+pub trait BencodeValueBufConstraint: std::hash::Hash + Eq + ByteBufT {}
 
+impl<T> BencodeValueBufConstraint for T where T: std::hash::Hash + Eq + ByteBufT {}
+
+/// A dynamic value when we don't know exactly what we are deserializing.
+/// Useful for debugging.
 #[derive(PartialEq, Eq, Clone)]
-pub enum BencodeValue<BufT: std::hash::Hash + Eq> {
+pub enum BencodeValue<BufT: BencodeValueBufConstraint> {
     Bytes(BufT),
     Integer(i64),
     List(Vec<BencodeValue<BufT>>),
     Dict(HashMap<BufT, BencodeValue<BufT>>),
 }
 
-impl<BufT: std::fmt::Debug + std::hash::Hash + Eq> std::fmt::Debug for BencodeValue<BufT> {
+impl<BufT> BencodeValue<BufT>
+where
+    BufT: BencodeValueBufConstraint + serde::Serialize,
+{
+    pub fn to_bytes(&self) -> Result<Vec<u8>, SerError> {
+        let mut bytes = vec![];
+        bencode_serialize_to_writer(self, &mut bytes)?;
+        Ok(bytes)
+    }
+}
+
+impl<BufT: std::fmt::Debug + BencodeValueBufConstraint> std::fmt::Debug for BencodeValue<BufT> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             BencodeValue::Bytes(b) => std::fmt::Debug::fmt(b, f),
@@ -117,8 +131,8 @@ impl<BufT: std::fmt::Debug + std::hash::Hash + Eq> std::fmt::Debug for BencodeVa
 
 impl<BufT> CloneToOwned for BencodeValue<BufT>
 where
-    BufT: CloneToOwned + std::hash::Hash + Eq,
-    <BufT as CloneToOwned>::Target: Eq + std::hash::Hash,
+    BufT: CloneToOwned + BencodeValueBufConstraint,
+    <BufT as CloneToOwned>::Target: BencodeValueBufConstraint,
 {
     type Target = BencodeValue<<BufT as CloneToOwned>::Target>;
 
@@ -160,20 +174,21 @@ mod tests {
 
     #[test]
     fn test_serialize_torrent_dyn() {
-        let mut buf = Vec::new();
+        let mut file_buf = Vec::new();
         let filename = "../librqbit/resources/ubuntu-21.04-desktop-amd64.iso.torrent";
         std::fs::File::open(filename)
             .unwrap()
-            .read_to_end(&mut buf)
+            .read_to_end(&mut file_buf)
             .unwrap();
 
-        let torrent: BencodeValueBorrowed = from_bytes(&buf).unwrap();
+        let torrent: BencodeValueBorrowed = from_bytes(&file_buf).unwrap();
 
-        let mut buf = Vec::<u8>::new();
-        bencode_serialize_to_writer(&torrent, &mut buf).unwrap();
+        let mut ser_buf = Vec::<u8>::new();
+        bencode_serialize_to_writer(&torrent, &mut ser_buf).unwrap();
 
-        let new_torrent = from_bytes(&buf).unwrap();
+        let new_torrent = from_bytes(&ser_buf).unwrap();
         assert_eq!(torrent, new_torrent);
+        assert_eq!(ser_buf, file_buf);
     }
 
     #[test]
@@ -191,5 +206,25 @@ mod tests {
         let mut buf = Vec::<u8>::new();
         bencode_serialize_to_writer(&test, &mut buf).unwrap();
         assert_eq!(&buf, b"d2:f1i100ee");
+    }
+
+    #[test]
+    fn test_dict_ordering() -> anyhow::Result<()> {
+        let hash_map = HashMap::from_iter((0..1000).map(|x| {
+            (
+                ByteString::from(x.to_string().into_bytes()),
+                BencodeValue::Integer(x),
+            )
+        }));
+        dbg!(&hash_map);
+        let orig = BencodeValue::Dict(hash_map);
+        dbg!(&orig);
+        let first_buf: ByteString = orig.to_bytes()?.into();
+        dbg!(&first_buf);
+        let first_deser = from_bytes(&first_buf)?;
+        assert_eq!(orig, first_deser);
+        let second_buf = first_deser.to_bytes()?.into();
+        assert_eq!(first_buf, second_buf);
+        Ok(())
     }
 }
--- a/crates/bencode/src/lib.rs
+++ b/crates/bencode/src/lib.rs
@@ -1,9 +1,18 @@
 mod bencode_value;
+mod raw_value;
 mod serde_bencode_de;
 mod serde_bencode_ser;
 
 pub use bencode_value::*;
+pub use raw_value::*;
 pub use serde_bencode_de::*;
 pub use serde_bencode_ser::*;
+
+use std::collections::BTreeMap;
+use std::fmt::Formatter;
+
+use serde::de::{DeserializeOwned, Error as _, Visitor};
+use serde::ser::{Impossible, SerializeStruct};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 pub use buffers::{ByteBuf, ByteString};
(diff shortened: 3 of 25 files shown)


anacrolix:
I found something like the raw value that could work, https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html. I imagine it could be done with some kind of reliance on ByteBuf.

ikatson:
I can take a closer look later (need to see locally with and IDE), but high-level comments so far.

First of all, thanks for the contribution. Removing hacks is good, learning Rust along the way even better!

> I don't quite know enough Rust, but I think you either want to (in the same way as my implementation) store the info bytes directly in a metainfo so you can hash them at will (and do a subsequent deserialize if you want to turn them into an info you can work with), or as I've done here, I've stored info as a BencodeValue, which can be deterministically turned into bytes for hashing, or deserialized further into an Info.

Not exactly. There are 3 ways info_hash can be created for use in rqbit
1. Reading an existing .torrent file or its bytes - and the subsequent deserializer hack that you are trying to remove.
2. Reading an existing info_hash from a magnet link
3. create_torrent_file() function, that is only used in tests so far.

Let's focus on 1 and 2 as actual production usages. What rqbit cares about it:
1. frequent access to info_hash for all kinds of purposes
2. frequent access to info and fields in it like filenames, piece hashes etc.

What it does not care about:
1. serializing an already deserialized torrent
2. thus no need to preserve serialization order (implied from above)
3. thus no need to store fields that we don't care about, because we don't understand them.

> I did notice you use HashMap for BencodeValue::Dict. That shouldn't allow serializing back into a different order, but I couldn't trigger that behaviour with the unit test. I'm not sure how to best trigger this behaviour in Rust. The solution would be to have an ordered mapping type, or have Dict use Vec internally and expose some kind of mapping interface? BitTorrent does specify that dict keys must be ordered like memcmp (but not in the original BEP 3), so client support is flakey, and that's why you want to deserialize info into something that retains the original ordering, at least if you intend to generate infohashes.

If order is important BTreeMap should be used. As you can see, it was never important previously, as there's no need to serialize info.

> and that's why you want to deserialize info into something that retains the original ordering, at least if you intend to generate infohashes.

Another option (which I probably would prefer) is to deserialize into both:
- a custom struct that has all the fields we know how to work with
- something that lets you dump it back preserving the order. The easiest thing here is just plain bytes - not even BencodeValue.

So if I were to do anything about it, I would store 3 things:
1. The info_hash for frequent access. Computing it on every call is expensive and unnecessary - it never changes.
2. The TorrentMetaV1Info - for all the fields we care about.
3. The raw bytes of TorrentMetaV1Info - in case we ever need to restore the original in its pure form (e.g. if an extended peer is asking us to give them the content of the torrent info, this feature doesn't exist today).

> Moving the infohashing out of bencode moved the sha1w dep to librqbit_core. I didn't retain the features, I assume they aren't in use (except on the main bin perhaps?), but if it was I can put them back in mapping features to core instead.

Great, let it burn! The --sha1* features are from the old times when I was measuring perf and picking the best option. But the one used right now is the clear winner, so all this cruft can be removed from everywhere.

> There's no way to deserialize from BencodeValue to a Rust type. I'm not well versed in serde, would that require a custom deserializer? Instead I serialized back into bytes then ran the bytes deserializer. It should still be correct, unless the Dict nondeterministic ordering strikes. At least for infohashing, there's no intermediate allocation as it can serialize straight into the hasher.

There's no way to deserialize from BencodeValue to a Rust type: that would require a custom Serializer and Deserializer for BencodeValue, which although can be done, is overkill for the task at hand.

Instead I serialized back into bytes then ran the bytes deserializer: this is a fine "hack", but with above recommendation I don't think we need it.

> How does one deserialize into bytes with serde?

[Code](https://github.com/ikatson/rqbit/blob/main/crates/bencode/src/serde_bencode_ser.rs#L484).
E.g. you create a ```Vec<u8>``` (possibly wrap into ```BufWriter``` for performance), then call bencode_serialize_to_writer on it.
```Vec<u8>``` implements std::io::Write.

> If it was possible to have info: ByteBuf in the metainfo that could be optimal for hashing after deserializing. With the BencodeValue route I went, I had to use BencodeValueOwned: I couldn't work out the trait and lifetime requirements to have it be BencodeValue. Do you know why that might be? Perhaps knowing whether BencodeValue or a raw ByteBuf type could be used in metainfo to avoid allocations could decide the best choice?

It should be possible. Smth like like serde_json's RawValue which you pointed out can do this. But of course for rqbit's deserializer it would need to be implemented, we can't reuse serde_json as it is specific to that crate.

That would require adding other "hacks" to serializer and deserializer. In fact serde_json's "RawValue" is very similar to what we were doing with "info_hash" - just that our hack wasn't reusable and only served one purpose.

I would imagine it would be possible to write something like this:

```
struct WithRawValueAndHash<T, ByteBuf> {
    deserialized_value: T,
    raw_value: ByteBuf,
    hash: Id20
}

struct TorrentMetaV1<ByteBuf> {
    ...
    info: WithRawValueAndHash<TorrentMetaV1Info<ByteBuf>, ByteBuf>,
    ...
}
```

And then somehow (not sure how, need to try) to make it work automatically with BencodeSerializer and BencodeDeserializer in a similar way to serde_json::RawValue

> I tried to add a std::io::Write for any ISha1, but Rust didn't like that, so I made a helper type then noticed you had a private one in a function somewhere that was identical.

If you cleaning this up anyway, we can just get rid of ISha1 generic, all the related features and multiple implementations. We can even get rid of the whole sha1w crate, and just have one "Sha1" struct librqbit_core.

> I have been toying with the idea of a Rust implementation of my torrent client for a while, but there's a few rough spots in my knowledge of Rust. If it works for you'll I'll make some more contributions like this if you have the capacity for it.

Sounds good, happy to help!

**Review by ikatson — COMMENTED** [review]
Looked closer at the code with my IDE.

Just a couple comments on the code - minor.

My biggest question before looking at it was how did you manage to pull off replacing "info: TorrentMetaV1Info" with "info: BencodeValue" without refactoring a ton of code.

But I got the answer - the session used only TorrentMetaV1Info already, and you convert it in only one place

File: crates/librqbit_core/src/torrent_metainfo.rs [hunk 1]
```diff
@@ -62,8 +55,17 @@ impl<BufType> TorrentMetaV1<BufType> {
...
+
+    /// v1 called out, because v2 is a thing, and migration is a PITA.
+    pub fn v1_info_hash(&self) -> Id20 {
+        let mut h = sha1w::Sha1::new();
+        bencode::bencode_serialize_to_writer(&self.info, &mut h.write()).unwrap();
+        Id20::new(h.finish())
+    }
 }
 
+type RawInfo = bencode::BencodeValueOwned;
```
ikatson, line 67: Is it possible to make this generic over BufT?

Also if it's used once probably doesn't make sense to have a separate type for it.

i.e why not just

```
info: BencodeValue<BufT>
```

ikatson:
Overall, I don't hate it, but I don't love it. 

The little hack was pretty innocent but made things simpler I think - as it was all structured types without converting back and forth between bytes and serialization. 

I'm not sure it's worth it without creating a general abstraction like "WithRawValueAndHash" above. I'm not sure it's worth creating that abstraction either though - hard to tell without trying.

Let me know what you think and if you agree with anything I said :)

ikatson:
> I'm not sure it's worth it without creating a general abstraction like "WithRawValueAndHash" above. I'm not sure it's worth creating that abstraction either though - hard to tell without trying.

I just tried this in [test-bencode-ser](https://github.com/ikatson/rqbit/compare/main...test-bencode-ser) branch for fun. Although cool that it works, but I'm still not sure it's worth the complexity. Maybe it is.

anacrolix:
> Let's focus on 1 and 2 as actual production usages. What rqbit cares about it:
> 
> 1. frequent access to info_hash for all kinds of purposes
> 2. frequent access to info and fields in it like filenames, piece hashes etc.

Agreed. If it wasn't clear, I want to expose the info_hash by doing the hashing just once. I think a wrapper around metainfo parsing that returns the metainfo, the info, and the infohash.

> What it does not care about:
> 
> 1. serializing an already deserialized torrent
> 2. thus no need to preserve serialization order (implied from above)
> 3. thus no need to store fields that we don't care about, because we don't understand them.

I agree. I did 1 just as a hack. I think it's clear that probably extracting the bytes directly for hashing is optimal. Then the order never matters as you say.

> > I did notice you use HashMap for BencodeValue::Dict. That shouldn't allow serializing back into a different order, but I couldn't trigger that behaviour with the unit test. I'm not sure how to best trigger this behaviour in Rust. The solution would be to have an ordered mapping type, or have Dict use Vec internally and expose some kind of mapping interface? BitTorrent does specify that dict keys must be ordered like memcmp (but not in the original BEP 3), so client support is flakey, and that's why you want to deserialize info into something that retains the original ordering, at least if you intend to generate infohashes.
> 
> If order is important BTreeMap should be used. As you can see, it was never important previously, as there's no need to serialize info.

Turns out there was a BTreeMap in the serializer, which is why I couldn't trigger a bug with bencode generated from rqbit.

> 
> > and that's why you want to deserialize info into something that retains the original ordering, at least if you intend to generate infohashes.
> 
> Another option (which I probably would prefer) is to deserialize into both:
> 
> * a custom struct that has all the fields we know how to work with
> * something that lets you dump it back preserving the order. The easiest thing here is just plain bytes - not even BencodeValue.

I suspect you can't deserialize into both an info with borrowed byte buffers, and generate a reference to the region of bytes the info was deserialized from. At least not without some refcell magic or other shenanigans.

> 
> So if I were to do anything about it, I would store 3 things:
> 
> 1. The info_hash for frequent access. Computing it on every call is expensive and unnecessary - it never changes.
> 2. The TorrentMetaV1Info - for all the fields we care about.
> 3. The raw bytes of TorrentMetaV1Info - in case we ever need to restore the original in its pure form (e.g. if an extended peer is asking us to give them the content of the torrent info, this feature doesn't exist today).

3 is very forward thinking of you. I guess to that end, it would be okay to deserialize and end up with a copy of the bytes that the info was deserialized from. Whatever the deserializer read from probably isn't sticking around.
 
> [Code](https://github.com/ikatson/rqbit/blob/main/crates/bencode/src/serde_bencode_ser.rs#L484). E.g. you create a `Vec<u8>` (possibly wrap into `BufWriter` for performance), then call bencode_serialize_to_writer on it. `Vec<u8>` implements std::io::Write.

Ah no I meant like json's RawValue. Essentially type check the input and end up with a reference to it in the output value without doing any allocation.

> I would imagine it would be possible to write something like this:
> 
> ```
> struct WithRawValueAndHash<T, ByteBuf> {
>     deserialized_value: T,
>     raw_value: ByteBuf,
>     hash: Id20
> }
> 
> struct TorrentMetaV1<ByteBuf> {
>     ...
>     info: WithRawValueAndHash<TorrentMetaV1Info<ByteBuf>, ByteBuf>,
>     ...
> }
> ```
> 
> And then somehow (not sure how, need to try) to make it work automatically with BencodeSerializer and BencodeDeserializer in a similar way to serde_json::RawValue

An interesting challenge, but it leaves you with fields that irrelevant to a serializer. Hence why I think it's more accurate to have info_bytes inside the metainfo, and then deserialize that again to get the info (and vice versa to serialize a metainfo, you serialize the info then put it in the info field).

> Sounds good, happy to help!

Thanks!

crates/librqbit/src/session.rs
```diff
@@ -893,16 +892,16 @@ impl Session {
...
                             trackers.clone(),
                             announce_port,
                             opts.force_tracker_interval,
                         )?
                     };
 
                     (
-                        torrent.info_hash,
-                        torrent.info,
+                        torrent.v1_info_hash(),
```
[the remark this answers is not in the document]
anacrolix: I intend to store the info hash once and pass that around. If it's computed here more than once I'll fix that.

crates/librqbit_core/src/torrent_metainfo.rs
anacrolix: I couldn't get that to work!

anacrolix:
> > I'm not sure it's worth it without creating a general abstraction like "WithRawValueAndHash" above. I'm not sure it's worth creating that abstraction either though - hard to tell without trying.
> 
> I just tried this in [test-bencode-ser](https://github.com/ikatson/rqbit/compare/main...test-bencode-ser) branch for fun. Although cool that it works, but I'm still not sure it's worth the complexity. Maybe it is.

Nice. I think this approach might work but without creating the object too, for the reason I gave about not having stuff specific to serializer or deserializer. Essentially RawValue would let you walk across bencode values without doing any allocation.

ikatson:
> Turns out there was a BTreeMap in the serializer

Ah, right, I remember doing it now. I did it this way, so that it doesn't matter if the passed-in object is a HashMap or BTreeMap (cause we can't distinguish them in the serializer, I guess).

> I suspect you can't deserialize into both an info with borrowed byte buffers, and generate a reference to the region of bytes the info was deserialized from.

If you mean self-referential structs, then yes it's impossible to do with safe Rust.
E.g.

```
struct Torrent {
info_bytes: Vec<u8>,
info: TorrentMetaInfo<... this can't reference "info_bytes">
}
```

However, there are 2 things that could be done theoretically, I'm not saying we need it:

1. For borrowed structs (that reference the original bencode binary), it's not a problem at all, as you can have an &[u8] both in "info_bytes" and "info".
2. For owned structs, one could use smth like the "bytes" crate, which exposes a reference-counted blob of bytes that you can slice.

> I guess to that end, it would be okay to deserialize and end up with a copy of the bytes that the info was deserialized from. Whatever the deserializer read from probably isn't sticking around.

Either a copy (which isn't a huge problem), or smth like "bytes" above.

> Essentially type check the input and end up with a reference to it in the output value without doing any allocation.

That's why everything is generic over "BufT", so that you can have both owned and borrowed versions. But the owned one doesn't do reference-counted byte blobs.

> Essentially RawValue would let you walk across bencode values without doing any allocation.

You could even today, but we it will be useless cause we can't store it.

E.g if we stored &[u8] or Vec<u8> with metainfo bytes, we can deserialize it into a TorrentMetaV1Info<&'a [u8]> (it's called TorrentMetaV1InfoBorrowed). But then we can't store it anywhere as everything now has a lifetime. And it's an object that is very frequently consulted.

anacrolix:
> > I'm not sure it's worth it without creating a general abstraction like "WithRawValueAndHash" above. I'm not sure it's worth creating that abstraction either though - hard to tell without trying.
> 
> I just tried this in [test-bencode-ser](https://github.com/ikatson/rqbit/compare/main...test-bencode-ser) branch for fun. Although cool that it works, but I'm still not sure it's worth the complexity. Maybe it is.

Thanks for this. Your implementation is much better than JSON's, I don't know why theirs is so convoluted. I learned a lot from it.

anacrolix:
I'm leaning toward 3 but without holding an intermediate value. I will update the PR in a day or two.

ikatson:
Sounds good @anacrolix , I'll leave it up to you.

crates/librqbit_core/src/torrent_metainfo.rs
```diff
@@ -37,7 +43,10 @@ pub struct TorrentMetaV1<BufType> {
         skip_serializing_if = "Vec::is_empty"
     )]
     pub announce_list: Vec<Vec<BufType>>,
-    pub info: TorrentMetaV1Info<BufType>,
+    /// Unmodified bytes corresponding to the info value. Used for computing infohash and
+    /// communicated to peers via the metadata extension.
+    #[serde(borrow = "'a")]
+    pub info: Option<RawInfo<'a, BufType>>,
```
anacrolix, line 49: @ikatson I don't suppose you could check this branch out in your IDE and help me solve the single compile error I have left? I am trying to support deserializing from borrowed data, but I get a lifetime I just can't figure out for having the raw info bytes in this field. It the lifetime should be the deserializer lifetime, but I don't know how to express it.

anacrolix:
I made a good mess that I want to clean up when I get prove it will compile. I mentioned where I'm stuck above.

The other difficulty I can see is if you want to be backward compatible with session states from old versions, how would you go about ensuring that the info field can be deserialized from the form it's currently in on disk (base64 of JSON), or whatever it is, it should probably be ignored when it fails to deserialize, and then when the state is reserialized, include the info in its raw byte form. I got an error running with a session from an older version of rqbit, and the torrent didn't show up in the UI, but the app continued to work and when I added the torrent again it serialized it correctly. Not sure how you would want to handle upgrading from old session states, or if you would just flag it as a breaking change in the release notes.

ikatson:
A few things:

> I don't suppose you could check this branch out in your IDE and help me solve the single compile error I have left?

There you go https://github.com/ikatson/rqbit/commit/b345dc5933dcae0ffeb19ca5c97c9ab5dc47876e

> how would you go about ensuring that the info field can be deserialized from the form it's currently in on disk

It uses custom functions to serialize/deserialize today - serialize_torrent/deserialize_torrent. You can just update them to try both formats (new and old). Let's not break backwards compat for no good reason.

Will put some more comments into the PR itself

crates/bencode/Cargo.toml
```diff
@@ -10,15 +10,12 @@ readme = "README.md"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[features]
```
ikatson, line 13: please rebase on main

crates/bencode/Cargo.toml
```diff
@@ -10,15 +10,12 @@ readme = "README.md"
...
 serde = {version = "1", features=["derive"]}
 buffers = {path = "../buffers", package="librqbit-buffers", version = "2.2.1"}
 clone_to_owned = {path = "../clone_to_owned", package="librqbit-clone-to-owned", version = "2.2.1"}
 anyhow = "1"
-sha1w = {path="../sha1w", default-features=false, package="librqbit-sha1-wrapper", version="2.2.1"}
\ No newline at end of file
+serde_bytes = "0.11.14"
+
+[dev-dependencies]
+serde_json = "1.0.115"
```
ikatson, line 21: why is this needed?

crates/bencode/Cargo.toml
```diff
@@ -10,15 +10,12 @@ readme = "README.md"
...
-sha1-rust = ["sha1w/sha1-rust"]
-
 [dependencies]
 serde = {version = "1", features=["derive"]}
 buffers = {path = "../buffers", package="librqbit-buffers", version = "2.2.1"}
 clone_to_owned = {path = "../clone_to_owned", package="librqbit-clone-to-owned", version = "2.2.1"}
 anyhow = "1"
-sha1w = {path="../sha1w", default-features=false, package="librqbit-sha1-wrapper", version="2.2.1"}
\ No newline at end of file
+serde_bytes = "0.11.14"
```
ikatson, line 18: let's not use serde_bytes, but keep using buffers crate instead. As we have handy wrappers around [u8], Vec<u8> already, there's no need to use serde_bytes.

crates/bencode/src/bencode_value.rs
```diff
@@ -93,18 +93,32 @@ where
     }
 }
 
-// A dynamic value when we don't know exactly what we are deserializing.
-// Useful for debugging.
+pub trait BencodeValueBufConstraint: std::hash::Hash + Eq + ByteBufT {}
```
ikatson, line 96: I don't like this - it's easier to depend on std::hash::Hash + Eq in implementation of ByteBufT, so you'll just have: "pub enum BencodeValue<BufT: ByteBufT>" if needed.

But at the same time I'm not sure why change the explicit bounds in the first place - it just increases the surface aread of the PR needlessly.

crates/bencode/src/bencode_value.rs
```diff
@@ -93,18 +93,32 @@ where
...
     List(Vec<BencodeValue<BufT>>),
     Dict(HashMap<BufT, BencodeValue<BufT>>),
 }
 
-impl<BufT: std::fmt::Debug + std::hash::Hash + Eq> std::fmt::Debug for BencodeValue<BufT> {
+impl<BufT> BencodeValue<BufT>
+where
+    BufT: BencodeValueBufConstraint + serde::Serialize,
+{
+    pub fn to_bytes(&self) -> Result<Vec<u8>, SerError> {
```
ikatson, line 114: Is this needed anywhere? If not, let's remove.

crates/bencode/src/lib.rs
```diff
@@ -1,9 +1,36 @@
...
+
+use serde::{
+    de::{value::BorrowedBytesDeserializer, DeserializeOwned, Error as _, MapAccess, Visitor},
+    ser::Impossible,
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+
+use clone_to_owned::CloneToOwned;
+
+fn escape_bytes(bytes: &[u8]) -> String {
```
ikatson, line 27: why is this needed?

crates/bencode/src/serde_bencode_de.rs
```diff
@@ -1,79 +1,96 @@
...
     field_context: Vec<ByteBuf<'de>>,
     parsing_key: bool,
-
-    // This is a f**ing hack
-    pub is_torrent_info: bool,
-    pub torrent_info_digest: Option<[u8; 20]>,
 }
 
 impl<'de> BencodeDeserializer<'de> {
+    fn buf(&'de self) -> &'de [u8] {
```
ikatson, line 11: why change this? I didn't need to do it in my reference implementation.

The original code used one field - buf, now you have 2 of them + a method that you need not forget to call. But most importantly, it worked just fine, so why change it?

I suggest not changing things unrelated to the PR just because you don't like the code for some reason. It increases the surface area to review, makes the change questionable. In general if it ain't broken, I wouldn't change it.

ikatson (changes requested):

ikatson:
A similar thing was merged in recently, so this is no longer needed. We now store reference-counted bytes, both the full info, and the deserialized one

PR #100, opened 2024-03-28
Add .desktop file configured for torrents

kpcyrd:
The tauri generated .desktop file is not configured for any file types or protocols, with this file rqbit-desktop is going to be suggested for opening magnet links and .torrent files.

However, I noticed commandline arguments are currently not understood by rqbit-desktop, and opening two files results in two rqbit-desktop instances.

### DIFF
--- a/resources/rqbit-desktop.desktop
+++ b/resources/rqbit-desktop.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Categories=Network;FileTransfer;P2P
+Comment=rqbit torrent client
+Exec=rqbit-desktop %U
+Icon=rqbit-desktop
+Name=rqbit-desktop
+Terminal=false
+Type=Application
+MimeType=application/x-bittorrent;x-scheme-handler/magnet;


ikatson:
Isn't it useless without rqbit-desktop understanding command line arguments? It will ignore the file you're trying to open anyway, won't it?

kpcyrd:
True, I should probably look into how to accept command-line arguments as part of this PR. To you have any pointers?

ikatson:
Just accepting isn't enough, you need to do smth with them, which isn't very trivial either:

- if it's not open yet, you need to start with the torrent that was passed in
- if it's already open (how to determine that?), you need to tell it somehow what to download

An ugly hack would be to assume that TCP API is on, in which case you can use that to
- test if it's running alreay or not
- submit a file for download

So if you are ready to tackle this, then this is the file to add this to (including parsing command line args): https://github.com/ikatson/rqbit/blob/main/desktop/src-tauri/src/main.rs

ikatson:
Also if an existing window is there would be cool to focus on it

ikatson:
The HTTP API client is in librqbit::http_api_client btw

ikatson:
Actually, thinking more about it, it's not enough - you don't want to simply add the file!
You want instead of open the dialog in the UI - the one that is opened when you click the UI buttons to add magnet or local file.
Which makes it not possible to do via existing HTTP API

PR #111, opened 2024-04-04
Add build steps of desktop application to README.md

ParadoxZero:
Added details on how to build the desktop app for rqbit manually to the sub directory's README.md file. There is some value to have the build details present along with the code and it's config.

### DIFF
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -1,7 +1,33 @@
-# Tauri + React + Typescript
+# rqbit GUI Application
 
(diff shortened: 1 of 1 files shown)


ParadoxZero:
This is probably not needed since you have mentioned the steps in DEV-GUIDE.md
I personally didn't notice the file. When I was exploring so had to figure out the steps before realizing you have created a cool makefile to automate all of this.

PR #129, opened 2024-04-30
Dynamic configurable storages

ikatson:


cocool97:
Hey, I took a look at the code and this looks promising ! 
From what I see, I've got two questions : 

- In `custom_storage.rs`, the storage factory is related to a torrent, that may contain many files. How can I set a storage for a given file in a torrent ? 
- Is is possible to implement the `Read` and `Seek` traits on a torrent file ? This would be useful and interoperable with "classic" Rust code (like web servers in my case).

Thanks for your time !

ikatson:
@cocool97 

> In custom_storage.rs, the storage factory is related to a torrent, that may contain many files. How can I set a storage for a given file in a torrent ?

you can implement a storage that does something different per file.

```
pub trait TorrentStorage: Send + Sync {
    // note the file_id. You can do whatever you want here
    fn pread_exact(&self, file_id: usize, offset: u64, buf: &mut [u8]) -> anyhow::Result<()>;
    // same here
    fn pwrite_all(&self, file_id: usize, offset: u64, buf: &[u8]) -> anyhow::Result<()>;
```

> Is is possible to implement the Read and Seek traits on a torrent file ? This would be useful and interoperable with "classic" Rust code (like web servers in my case).

Look at streaming.rs - it's already there, but it's AsyncRead and AsyncSeek (not sync, but async).
There's also a web server (axum) [endpoint](https://github.com/ikatson/rqbit/blob/main/crates/librqbit/src/http_api.rs#L160) that uses it to stream it over HTTP. This is merged into main already, it's only the dynamic storages part that isn't yet.

cocool97:
Ok, this is indeed possible. But wouldn't it be easier for every user of this library to set a "per-file in torrent" storage ? 
Keeping it on a torrent basis would force everyone to implement the per-file logic.

ikatson:
Can you provide an API example that you see in mind? I'm also clueless as to why someone would need smth like that - a separate storage per file in torrent, so would appreciate if you highlight how do you intend to use it.

Looking at it, @anacrolix /torrent seems to have per torrent flat storage (not even per file). Although I also don't know what's the use of storages, but it was very handy today to debug a slow disk issue.

ikatson:
I'm tired of rebasing this, so merging into main

PR #152, opened 2024-07-20
Add redirect for /web

izderadicka:
Small one -  I typed http://localhost:3030/web   in browser (without training /).
Took me while before I realize my mistake - hope it can help some other distracted user.


ikatson:
Thanks for the fix!

Issue #153, opened 2024-07-20
Calling blocking function in streaming?


izderadicka:
Hi,

first thanks for great SW - I just started to play with it, but works fine, is simple and ergonomic.

This is rather a question  - I was looking into implementation of file streaming - e.g.  `FileStream` in `rqbit/crates/librqbit/src/torrent_state/streaming.rs`.

Filestream implements tokio's AsyncRead trait,  and and within poll_read it calls this:

```rust
poll_try_io!(poll_try_io!(self.torrent.with_storage_and_file(
            self.file_id,
            |files, _fi| {
                files.pread_exact(self.file_id, self.position, buf)?;
                Ok::<_, anyhow::Error>(())
            }
        )));
```

Looking into code for `files.pread_exact` - it looks like this is blocking call for default FilesystemStorage calling blocking File IO from std::io.   
Is this problem or am I missing something?

PR #155, opened 2024-07-21
Other way to handle blocking in streaming.rs

izderadicka:
Hi,

this relates to #153 .

`block_in_place` - I've tried here - it's relatively straightforward -  https://github.com/ikatson/rqbit/compare/main...izderadicka:naive_streaming_fix  

I also found a problem on e2e_streaming test - at least on my linux - it tried to create directory  in root /does-not-matter, which failed due to permissions..

Here I tried approach with `spawn_blocking` - which I feel is more "appropriate".  With blocking in event loop I think there can be problems if tokio runs in current thread executor.   On the other hand I understand that reading from local file would be fairly quick (taking into account Linux kernel's  fs caches etc.).  So it might not be faster, as some additional allocation is involved, when task should run in different thread - this might be still bit optimized - here is first rough attempt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 .DS_Store
+/.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,0 @@
-{
-  "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "[javascript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "[rust]": {
-    "editor.defaultFormatter": "rust-lang.rust-analyzer"
-  }
-}
--- a/crates/librqbit/src/http_api.rs
+++ b/crates/librqbit/src/http_api.rs
@@ -170,7 +170,7 @@ impl HttpApi {
             let range_header = headers.get(http::header::RANGE);
             trace!(torrent_id=idx, file_id=file_id, range=?range_header, "request for HTTP stream");
 
-            if let Some(range) = headers.get(http::header::RANGE) {
+            if let Some(range) = range_header {
                 let offset: Option<u64> = range
                     .to_str()
                     .ok()
--- a/crates/librqbit/src/tests/e2e_stream.rs
+++ b/crates/librqbit/src/tests/e2e_stream.rs
@@ -101,6 +101,8 @@ async fn e2e_stream() -> anyhow::Result<()> {
     let mut buf = Vec::<u8>::with_capacity(8192);
     stream.read_to_end(&mut buf).await?;
 
+    assert_eq!(buf.len(), orig_content.len(), "sizes differ");
+
     if buf != orig_content {
         panic!("contents differ")
     }
--- a/crates/librqbit/src/tests/e2e_stream.rs
+++ b/crates/librqbit/src/tests/e2e_stream.rs
@@ -1,6 +1,7 @@
 use std::{net::SocketAddr, time::Duration};
 
 use anyhow::Context;
+use tempfile::TempDir;
 use tokio::{io::AsyncReadExt, time::timeout};
 use tracing::info;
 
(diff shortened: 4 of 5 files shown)


izderadicka:
Hi,
thanks for comments,  I put back `register_waker` in the callback, I did not realize it might be an issue, but now I think I understand the pattern.

Concerning the overall usability of the change, I'm not sure.  I have done some quick dirty measurements (streaming down  1GB video with curl) and have not seen any notable performance difference.

The only difference is if `block_in_place` is used as simple workaround  (as mentioned above) this indeed does not work in current thread executor (-s option).   This proposed fix in current thread executor behaves OK, same as original code.

The code of this PR is more complex, that the feature of async code - I think there is no other way around, if you're composing futures "manually" you have do do all the low level stuff - polling etc. 

On the other hand, it can be useful for future -  I think that ideally Storage should be async trait ( which I think got stabilized, but I think there is still issue with async trait being a trait object - but this also will be solved in future). If storage will be async for reading then the streaming should  be  similar to proposed change.

So I leave it to your discretion, if this PR could be useful now.

ikatson:
Looking at this again closer, I'm still hesitant to merge it because of unnecessary (at least at this time while Storage isn't async) complexity. 

Dealing with the buffer / position etc was a bit tricky already, but this brings it to a whole next level - there's a lot of control flow and branching and just plain more code. Looking at it, it all seems reasonable, most of the complexity comes from writing Poll implementations manually being painful (albeit an enjoyable brain exercise).

As far as I understand, this is the main argument:

> The only difference is if block_in_place is used as simple workaround (as mentioned above) this indeed does not work in current thread executor (-s option). 

I should have expanded on it, but this was already encountered and worked around previously elsewhere in the codebase. You could just run the callback straight (without block_in_place) in single-thread mode - this was done already in the past a few times hence the BlockingSpawner helper in spawn_utils.rs. The spawner already gets created at session initialization, and then you can just copy it into the stream from ManagedTorrentInfo.spawner. Then instead of calling tokio::block_in_place you call spawner.spawn_block_in_place.

> I think that ideally Storage should be async trait

That would be nice to try, but I fear it'll also make many things quite a bit more complicated (as this PR shows) and [slower](https://github.com/tokio-rs/tokio/issues/3664).

Hence I'm leaning towards keeping it simple and performant (both streaming.rs and not making storage async) with a couple strategic workarounds until (if ever) it becomes an issue.

izderadicka:
Hi,
I do understand your arguments,  few comments:
- async is slower - I'd say it depends - I have done some toy benchmarks (mentioned above) and has not seen any slowdown. Even your mentioned issue in Tokio - they finally have been able to achieve good performance, when utilizing more parallelization in Tokio.  But true is that difference was small, so it probably does not justify the effort :-)
- I'm not so big expert on system programing, so I do not have that gut feeling about what potentially blocking call is still OK in async context - so I tend to follow the "formal" rules here - that's why I rose this issue - now I'm bit more enlighten and thanks for your patience and explanations 

So what I propose I will open new PR - with BlockingSpawner (I already saw this one) and fixed test - because that one was broken on my Linux box. I already have branch for it.  And this one can stay parked for some future days. Agreed?

ikatson:
Thanks @izderadicka for your patience! This sounds good.

As to async file IO being slower: if you look what it's doing [under the hood](https://github.com/tokio-rs/tokio/blob/master/tokio/src/fs/file.rs#L583), it's plain more code (instructions, allocations, inter-thread communication) than one syscall in blocking mode.

So taken in isolation it will be slower by definition unless you use smth like io-uring.

Of course, if it blocks the executor and other unrelated tasks, this isn't great and may slow down the application as a whole. But in reality, if those reads are small, the disk is fast, and we workaround it using block_in_place to mark those places, it should be ok.

I'd say complexity is my primary concern, performance secondary, but still important.

izderadicka:
I'm closing this one - see at #157 instead.  This one might be useful in some future, when storage will become async.

PR #156, opened 2024-07-25
Add support for TB and PB in formatBytes in webui

kielerdotdev:


ikatson:
Thank you! Can you please run also "make webui-build" so that it's embedded in the binary

ikatson:
Actually whatever I'll do it when I release

PR #162, opened 2024-08-02
Add content type to streams

izderadicka:
This change is motivated by need to stream/open/download other files from torrent than mp4/mkv/avi.
Usecase is mainly if rqbit is running on other computer (like nas server) - so I can see what is in other files, download them to notebook etc.  Ad there are other media files, which streams nicely in browser - mp3 ....

Changes:
- add content-type header to stream - helps with handling the file in browser
- change slightly when content-length is send - always when there is not  content-range
- in UI - add link to all files in torrent and add target to link to open in new tab

Note: for some reason  files `crates/librqbit/webui/dist`, where checked in to repo, but they are also in .gitignore - I guess they are result of your release build process - so I did not included them.

### DIFF
--- a/crates/librqbit/src/api.rs
+++ b/crates/librqbit/src/api.rs
@@ -331,23 +331,18 @@ fn torrent_file_mime_type(
     info: &TorrentMetaV1Info<ByteBufOwned>,
     file_idx: usize,
 ) -> Result<&'static str> {
-    let file_name = info
-        .iter_filenames_and_lengths()?
-        .enumerate()
-        .find_map(|(idx, (f, _))| {
-            if idx == file_idx {
-                f.iter_components()
-                    .last()
-                    .and_then(|r| r.ok())
-                    .and_then(|s| mime_guess::from_path(s).first_raw())
-            } else {
-                None
-            }
-        });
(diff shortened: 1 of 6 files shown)


crates/librqbit/src/api.rs
```diff
@@ -320,3 +326,28 @@ fn make_torrent_details(
...
     })
 }
+
+fn torrent_file_mime_type(
+    info: &TorrentMetaV1Info<ByteBufOwned>,
+    file_idx: usize,
+) -> Result<&'static str> {
+    let file_name = info
+        .iter_filenames_and_lengths()?
+        .enumerate()
```
ikatson, line 336: Rather then enumerate().find_map(if ...) wouldn't just `nth(file_idx)` work?

ikatson (commented):
Thanks, this looks beautiful and well reasoned! There's a small nit I left, otherwise LGTM

ikatson (approved):

crates/librqbit/src/api.rs
izderadicka: Absolutely,  bad me - cannot remember all those cool iterators combinators - fixed in next commit.

PR #167, opened 2024-08-07
Playlis for playable media

izderadicka:
Rationale behind this PR:

Many torrents contains several playable files (series, albums ...). 
Good players like VLC has posibility to open a playlist - combined with streaming capability of rqbit this enables to play whole content of a torrent in the player (just by open playlist link in the player).

This is a minimal implementation, that works with VLC, couple more things I thought about:
- MIME type of playlist - the only official MIME is registered by Apple for HLS - this is not exactly this case as as text/plain works with VLC, so  I did not explore further - this  article is good source for m3u playlist "standard" - https://en.wikipedia.org/wiki/M3U#Internet_media_types
- ordering of files - now using the ordering provided by the program, not sure if reodering could be useful.
- UI icon - now on whole torrent - but I though also on possibility to have it on directory level. Current implementation was easiest to implement. But possibility to have playlist for  just particular directory sounds interesting.

### DIFF
--- a/crates/librqbit/src/http_api.rs
+++ b/crates/librqbit/src/http_api.rs
@@ -1,6 +1,6 @@
 use anyhow::Context;
 use axum::body::Bytes;
-use axum::extract::{Path, Query, State};
+use axum::extract::{Path, Query, Request, State};
 use axum::response::{IntoResponse, Redirect};
 use axum::routing::{get, post};
 use futures::future::BoxFuture;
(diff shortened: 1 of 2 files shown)


ikatson:
Code and feature LGTM, thanks for this!

A couple questions / suggestions (can all be done in a separate PR):
1. In the UI, looks like you added the icon just next to "delete". There is a dropdown that opens when clicking the cog, maybe move it somewhere there under the dropdown not to clutter the UI with buttons?
2. Would be cool to use the react modal instead of alert(). There's a few uses of them throughout already (including components), should be easy to add.
3. In the modal would be cool to display a copyable link, maybe even put it into clipboard
4. Can you add a screenshot how it looks in the UI pls so that I don't need to compile and check

ikatson:
Addresed these here https://github.com/ikatson/rqbit/pull/169

PR #171, opened 2024-08-09
Sort playlist and fix playlist URL copy

izderadicka:
Thanks for merging #167 and fixing UI in #169.  Couple additions:

Playlist should be sorted  and  copied playlist URL should be always absolute (which was not the case, when Web was served from localhost:3030 - then apiUrl is "").


ikatson:
Thanks!

PR #173, opened 2024-08-09
Playlist ordering and  mime

izderadicka:
Thanks for #172 - definitely better code organization,  just couple of edits from my linux perspective:
- torrent PL should be sorted by filename, torrent file id ordering is often not "natural"
- apple MIME is not working well with VLC on linux - so leaving it for apple platforms,  simple text/plain works on linux fine.

### DIFF
--- a/crates/librqbit/src/http_api.rs
+++ b/crates/librqbit/src/http_api.rs
@@ -183,6 +183,7 @@ impl HttpApi {
                         "application/vnd.apple.mpegurl; charset=utf-8",
                     )]
                 } else {
+                    // apple mime does not work with VLC on linux
                     [("Content-Type", "text/plain; charset=utf-8")]
                 },
                 body,
--- a/crates/librqbit/src/http_api.rs
+++ b/crates/librqbit/src/http_api.rs
@@ -152,7 +152,7 @@ impl HttpApi {
                     }
                 })
                 .collect::<Vec<_>>();
-            playlist_items.sort();
+            playlist_items.sort_by(|left, right| left.1.cmp(&right.1));
             Ok(playlist_items)
         }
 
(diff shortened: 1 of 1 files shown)


izderadicka:
On [b38c371](https://github.com/ikatson/rqbit/pull/173/commits/b38c371d3f5070969757c60cf63b993cca9b84ea) test failure was caused by platform variations?  Same code, except comment in next commit.

ikatson:
The tests flake out sometimes, must be a bug somewhere.

Your code style was good btw, I only split it up to reuse parts of your function.

Btw I tried pasting the playlist URL into VLC and it doesn't seem to work: it seems VLC expects playlist to be a local file, and when pasting into "open network" (or whatever it's called) nothing happens. Only if I download the playlist, then open it works.

Is it the same for you or I'm doing something wrong?

izderadicka:
@ikatson  Opening playlist URL in VLC works for me on linux, if MIME is text/plain  - I think VLC has problem with apple MIME type - application/vnd.apple.mpegurl - that's why I  left it there only for Apple platforms.  This MIME is officially for Apple HLS - and this is not our use case. But I thought, as you added this MIME, it might have some advantage on Apple?

Or with this MIME, it might require extended M3U syntax -  e.g. start with #M3UEXT line etc.  Will try

ikatson:
> Opening playlist URL in VLC works for me on linux, if MIME is text/plain

Thanks, that explains why nothing actually worked for me on Mac! I added MIME type there not because it wasn't working without it, but just because the [wikipedia page for M3U](https://en.wikipedia.org/wiki/M3U#Internet_media_types) was listing this mime type on the first line.

It actually says later that that apple one is for HLS (and I didn't read that at first), and a better one to use for general M3U is e.g. "application/mpegurl".

Fixing this here: https://github.com/ikatson/rqbit/pull/176

PR #174, opened 2024-08-10
Remove ignored files

izderadicka:
These files are ignored by git, but were pushed into repo in past.
They just mess with newly generated UI files.

Used :
```
git ls-files -i -c --exclude-standard | xargs git rm --cached
```

to get rid of them  - I guess it could be useful also for main branch.


ikatson:
As you can see, these files are needed for the build.
They are embedded into the binary so that /web/ URL works. They are in .gitignore so that no new generated files get added, only the ones explicitly added previously.

I agree they are annoying though. A better solution might be to run "npm build" as part of "build.rs".

izderadicka:
Thanks for explanation -I realized that rust build was problem, when I saw all jobs failed.
Just one note - adding them to .gitignore does not help,   as long as they are in repo, they will always be shown as modified, after `npm run build` -  
```
$ git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   crates/librqbit/webui/dist/assets/index.css
        modified:   crates/librqbit/webui/dist/assets/index.js
        modified:   crates/librqbit/webui/dist/manifest.json
        modified:   crates/librqbit/webui/node_modules/.package-lock.json

no changes added to commit (use "git add" and/or "git commit -a")
``` 

No big issue - just requires bit more attention to what is included in commits.

ikatson:
Yes that's expected. These files don't get generated during rust build, that's why they are committed.

But as I mentioned above, I think we need to get rid of them if favor of building npm in build.rs, cause these files are causing a bit of pain

ikatson:
Addressed in #180

Issue #170, opened 2024-08-08
Implement hash-based API for integration with Flood UI


APT37:
I appreciate the built-in UI for its sleekness, but it's not advanced enough (yet). Implementing a hash-based API would enable integration with Flood.

Here is the corresponding issue over at Flood: https://github.com/jesec/flood/discussions/769

ikatson:
Addressed in #187

PR #187, opened 2024-08-15
Hash-based API in addition to integer based

ikatson:


ikatson:
This fixes https://github.com/ikatson/rqbit/issues/170

PR #206, opened 2024-08-21
Fill yourip in extended handshake

izderadicka:
From what I see yourip is every extended handshake - can be filled easily on our side with something like this PR.


ikatson (approved):
Thank you!

PR #210, opened 2024-08-24
WIP: PEX - Peer Exchange

izderadicka:
This is WORK IN PROGRESS (WIP)  on adding PEX.

I wanted to share early - to get some feedback if I'm moving in wrong direction.

If this "early draft PR" does not work with project workflow,   sorry for that, just close it.
WIP PRs did work for me to have place to discuss and share on particular feature, but if this does not fit this project sorry again. 

I thought that PEX could be usefull, and should not be so difficult at least on receiving side. And I can try something bit more substantial ....  But  I do not have much time left - so it can take quite some time to finish.


ikatson:
On a quick look from phone, seems like a reasonable thing to have. Code is mostly ok, minor comments so far not worth writing them down until this is finished. 

Thanks for the early feedback ask, so far so good

crates/peer_binary_protocol/src/extended/ut_pex.rs
```diff
@@ -0,0 +1,143 @@
...
+            added_f: self.added_f.clone_to_owned(within_buffer),
+            added6: self.added6.clone_to_owned(within_buffer),
+            added6_f: self.added6_f.clone_to_owned(within_buffer),
+            dropped: self.dropped.clone_to_owned(within_buffer),
+            dropped6: self.dropped6.clone_to_owned(within_buffer),
+        }
+    }
+}
+
+pub struct PeerAddrIterator<'a> {
```
ikatson, line 73: You probably don't need a custom iterator. You can iterate bufs with .chunks() and then .map() the resulting subslices

crates/peer_binary_protocol/src/extended/ut_pex.rs
```diff
@@ -0,0 +1,143 @@
...
+
+
+    }
+}
+
+impl<B> UtPex<B>
+where
+    B: AsRef<[u8]>,
+{
+    pub fn added_peers<'a>(&'a self) -> anyhow::Result<Box<dyn Iterator<Item = PexPeerInfo> + 'a>> {
```
ikatson, line 100: I'd avoid allocation here as it's quite easy to do.

two options: 
1. either use itertools::Either (minimal changes to your existing code, good enough)
2. Option is also an iterator, and you can flatten it.

Option 2 would be smth like this:

```
if self.added.as_ref().map(|added| added.len() % 6 != 0).unwrap_or(false) {
    bail!(...)
}

let ips = self.added.as_ref().map(|addrs| addrs.chunk_exact(6)).flatten();
let flags = self.added_f.as_ref().map(|flags| flags.iter().copied()).flatten();

return ips.zip(flags).filter_map(|(ip, flags)| PexPeerInfo::from_bytes(ip, flags).ok())

```

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1657,4 +1671,21 @@ impl PeerHandler {
...
         Ok(())
     }
+
+    fn on_pex_message<B>(&self, msg: UtPex<B>)
+    where
+        B: AsRef<[u8]> + std::fmt::Debug,
+    {
+        // TODO: this is just first attempt at pex - will need more sophistication on adding peers - BEP 40,  check number of live, seen peers ...
+        if let Ok(peers) = msg.added_peers() {
+            peers.for_each(|peer| {
```
izderadicka, line 1681: Clearly adding eagerly is not good  - it takes huge amount of CPU - will need to find something better

ikatson: What eats CPU exactly out of this?

ikatson: Adding eagerly should be the way to go, just need to fix our workaround the issues

ikatson:
I tried to look into it, and it seems totally ok to me. The peers get added, I don't see why this in particular would increase CPU usage.
Tweaked it all to fix all the things I mentioned + nicer debug

https://github.com/izderadicka/rqbit/pull/1

ikatson:
Also profiled in release mode, no CPU issues noticed, works like a charm. On my box, the absolute majority of the CPU is taken by:
- syscal write to disk (45%)
- read from network (14%)
- checksum (9%)
- write to network (4%)

The rest is rqbit itself + tokio overhead.

crates/peer_binary_protocol/src/extended/ut_pex.rs
izderadicka: added flags - added_f - are optional - so we can have addresses  but not flags, in that case zip will not work. But chunks  might make things easier - will look into it.

ikatson: Look into the PR I put into your repo I've done it already

izderadicka:
Still see about 150% CPU -  it's on  big swarm - https://releases.ubuntu.com/24.04/ubuntu-24.04-desktop-amd64.iso.torrent?_gl=1*1bi12tz*_gcl_au*MTA4NjA0NjI5Ny4xNzI0NDkzMjYy&_ga=2.207753554.1199975540.1724493262-752598451.1714415561

ikatson:
Just checking, can that be that with that branch there's many more live peers that you observe?

The CPU would be proportional to the number of live (i.e. connected and communicating) peers usually. 

Otherwise maybe you can profile with perf or whatever you have on your system to see where's the new hotspot? I didn't see any on the torrent I checked

ikatson:
I'll check that Ubuntu one too

ikatson:
Just tried the ubuntu one.

System: macbook air m1
Max CPU usage - 80%
Seen peers - about 3000

There are 0 hotspots in the new code, all of "on_received_message" is in "on_received_piece".
Screenshot 2024-08-25 at 21 58 43

1. Can you ensure you're profiling a release build
2. Ensure you don't write logs / tracing is at "info" level.

You can use the newly added "make devserver-profile" command to do all that.

ikatson:
@izderadicka any updates on this? I think I can just go ahead and merge this + https://github.com/izderadicka/rqbit/pull/1, as it seems useful on its own already for discovering peers even if we don't send our peers yet

ikatson:
> Also it now looks only for IPv4 added peers - should look also at other categories.

I addresed this already.

Let's do this: I merge this as-is plus my changes on top. If we figure out perf issues we fix them, but I don't see any so far as I mentioned. The PR is useful already on its own and it doesn't need to wait until all PEX features are implemented.

I'll merge this one instead https://github.com/ikatson/rqbit/pull/221 which includes your changes

Issue #211, opened 2024-08-25
Torrent does not download when removing read block from cache


cocool97:
I'm trying to implement a custom `TorrentStorage` where pieces would be remove from this storage when being read by a client, to prevent much memory from being used.

The implementation is quite simple, and does not change much from your custom `InMemoryStorageExample` implementation.

When I keep all pieces in storage, everything works fine, but obviously memory goes up. With the following code, torrent does not download at all :

```rust
fn pread_exact(&self, file_id: usize, offset: u64, buf: &mut [u8]) -> anyhow::Result<()> {
        log::debug!("pread_exact {file_id} {offset}");
        let fi = &self.file_infos[file_id];
        let abs_offset = fi.offset_in_torrent + offset;
        let piece_id: u32 = (abs_offset / self.lengths.default_piece_length() as u64).try_into()?;
        let piece_offset: usize =
            (abs_offset % self.lengths.default_piece_length() as u64).try_into()?;
        let piece_id = self.lengths.validate_piece_index(piece_id).context("bug")?;

        let mut g = self.map.write();
        // Get and remove this data from buffer to free space
        let inmp = g.remove(&piece_id).context("piece expired")?;
        buf.copy_from_slice(&inmp[piece_offset..(piece_offset + buf.len())]);

        Ok(())
    }
```

Are you for example reading the Storage when a block has been added using `pwrite_all` ? This would explain the issue I have as block would be remove directly after being downloaded. 

I'm using `rqbit` vendored from git directly, with latest version `v7.0.0-beta3`

cocool97:
How can it be done programmatically ?
I don't think that I have control over it in the `TorrentStorage` trait, and the goal is to not have everything in RAM at the end of the download

ikatson:
I'll be fine with a PR to add smth like "on_piece_completed" callback to TorrentStorage

cocool97:
Ok seems good yes, I'll take care of it ;)

PR #219, opened 2024-08-27
feat: add on_piece_completed method on TorrentStorage

cocool97:
As discussed in https://github.com/ikatson/rqbit/issues/211, a custom `TorrentStorage` that would remove blocks when read by `pread_exact()` is currently not working as this method is also called when checking torrent piece hashes.

This PR adds a new method on `TorrentStorage` trait, `on_piece_completed()` that is automatically called by rqbit when a piece has been downloaded and checked. This allows implementations to deal with this new information as they think is best.

`InMemoryExampleStorage` structure has been updated to remove blocks when they have been read by a stream.

For all other implementations this new method only returns Ok(()) as I don't think that this has to be handled explicitly.

**Note:** I think there is however still an issue but apart from this new method added as blocks are loop-downloaded as they are no more available in the `TorrentStorage`

### DIFF
--- a/crates/librqbit/src/storage/examples/inmemory.rs
+++ b/crates/librqbit/src/storage/examples/inmemory.rs
@@ -8,28 +8,50 @@ use crate::type_aliases::FileInfos;
 
 use crate::storage::{StorageFactory, StorageFactoryExt, TorrentStorage};
 
-struct InMemoryPiece {
-    bytes: Box<[u8]>,
+pub struct InMemoryPiece {
+    pub content: Box<[u8]>,
+    pub has_been_validated: bool,
 }
 
 impl InMemoryPiece {
-    fn new(l: &Lengths) -> Self {
+    pub fn new(l: &Lengths) -> Self {
         let v = vec![0; l.default_piece_length() as usize].into_boxed_slice();
-        Self { bytes: v }
+        Self {
+            content: v,
+            has_been_validated: false,
+        }
+    }
+
+    pub fn can_be_discard(&self, upper_bound_offset: usize) -> bool {
+        self.has_been_validated && upper_bound_offset >= self.content.len()
     }
 }
 
 #[derive(Default, Clone)]
-pub struct InMemoryExampleStorageFactory {}
+pub struct InMemoryExampleStorageFactory {
+    max_ram_size_per_torrent: usize,
+}
+
+impl InMemoryExampleStorageFactory {
+    pub fn new(max_ram_size_per_torrent: usize) -> Self {
+        Self {
+            max_ram_size_per_torrent,
+        }
+    }
+}
 
 impl StorageFactory for InMemoryExampleStorageFactory {
     type Storage = InMemoryExampleStorage;
 
     fn create(
         &self,
-        info: &crate::torrent_state::ManagedTorrentInfo,
+        info: &crate::torrent_state::ManagedTorrentShared,
     ) -> anyhow::Result<InMemoryExampleStorage> {
-        InMemoryExampleStorage::new(info.lengths, info.file_infos.clone())
+        InMemoryExampleStorage::new(
+            info.lengths,
+            info.file_infos.clone(),
+            self.max_ram_size_per_torrent,
+        )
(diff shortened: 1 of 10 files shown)


crates/librqbit/src/storage/examples/inmemory.rs
```diff
@@ -108,14 +144,29 @@ impl TorrentStorage for InMemoryExampleStorage {
...
-    fn init(&mut self, _meta: &crate::ManagedTorrentInfo) -> anyhow::Result<()> {
+    fn init(&mut self, _meta: &crate::ManagedTorrentShared) -> anyhow::Result<()> {
         Ok(())
     }
 
     fn remove_directory_if_empty(&self, _path: &Path) -> anyhow::Result<()> {
         Ok(())
     }
+
+    fn on_piece_completed(&self, file_id: usize, offset: u64) -> anyhow::Result<()> {
```
ikatson, line 159: it would make more sense to pass ValidatedPieceIndex to it rather than file_id and offset.
Piece can span multiple files, so file_id and offset are kind of meaningless.

crates/librqbit/src/storage/examples/inmemory.rs
```diff
@@ -41,10 +63,15 @@ pub struct InMemoryExampleStorage {
     lengths: Lengths,
     file_infos: FileInfos,
     map: RwLock<HashMap<ValidPieceIndex, InMemoryPiece>>,
+    max_ram_size_per_torrent: usize,
 }
 
 impl InMemoryExampleStorage {
-    fn new(lengths: Lengths, file_infos: FileInfos) -> anyhow::Result<Self> {
```
ikatson, line 47: please don't change the example storage. It's there just to show how to implement a custom storage and serves no other purpose.

Deleting pieces after completion is non conformant, and even if done needs to be done properly and probably would span other changes in librqbit which I'm not sure about at least yet.

crates/librqbit/src/storage/examples/mmap.rs
```diff
@@ -63,11 +63,15 @@ impl TorrentStorage for MmapStorage {
...
-    fn init(&mut self, _meta: &ManagedTorrentInfo) -> anyhow::Result<()> {
+    fn init(&mut self, _meta: &ManagedTorrentShared) -> anyhow::Result<()> {
         Ok(())
     }
 
     fn remove_directory_if_empty(&self, _path: &std::path::Path) -> anyhow::Result<()> {
         Ok(())
     }
+
+    fn on_piece_completed(&self, _file_id: usize, _offset: u64) -> anyhow::Result<()> {
```
ikatson, line 74: It would probably make sense to implement default in the trait itself not to add this to every single storage example that exists.

There's also "storage_middleware" feature that will break if turned on (although I do understand it will break already anyway because ManagedTorrentInfo was renamed, but that's a separate thing)

ikatson (changes requested):

crates/librqbit/src/storage/examples/inmemory.rs
cocool97: Done in 9183df0ebd14b3026028cda077b0757128facaed. This indeed makes more sense

crates/librqbit/src/storage/examples/mmap.rs
cocool97: Added default (empty) implementation in the trait itself, can still be override in trait implementations.

I checked and a `cargo check --all-features` build (apart from an error of multiple definition of Sha1

```
error[E0428]: the name `Sha1` is defined multiple times
  --> crates/sha1w/src/lib.rs:79:1
   |
76 | pub type Sha1 = crypto_hash_impl::Sha1CryptoHash;
   | ------------------------------------------------- previous definition of the type `Sha1` here
...
79 | pub type Sha1 = ring_impl::Sha1Ring;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Sha1` redefined here
   |
   = note: `Sha1` must be defined only once in the type namespace of this module
```

Issue #220, opened 2024-08-27
 No way to restrict uploading to peers


DoctorRyner:
How to make sure it'll not upload anything? I have really low upload speeds

ikatson:
> How to make sure it'll not upload anything?
There's no way at the moment. The best you can do atm is --disable-upnp which will not expose your IP to the internet. But it'll still upload to peers made with outgoing connections if they need pieces.

> I have really low upload speeds
If it uploads as fast as it can though, is it creating any issues? It's not obvious how is that a problem

DoctorRyner:
> If it uploads as fast as it can though, is it creating any issues? It's not obvious how is that a problem

That's the issue, I can't upload anything if I use torrent. I can't send messages/images properly in messaging apps, etc.

And I also live in Germany, if it turn out there something licensed in the torrent for some reason, like discontinued ISO files for Nintendo Wii games or just anything, I'm gonna get a fine of few thousands euro. We actually really have to use torrents carefully since there are layer companies sending those claims left and right on anything suspicious

ikatson:
While I can't comment on legal matters, it's important to use rqbit responsibly and in accordance with local laws.

However, I understand that the client using all available upload bandwidth can make it difficult for other apps to function properly. I'll consider adding a feature to disable uploading to help manage this.

DoctorRyner:
@ikatson or I can actually implement it for you if you give me some leads how your app handles uploading

ikatson:
Done this here https://github.com/ikatson/rqbit/pull/223, please validate if this works as expected

ikatson:
Done in #223, closing

DoctorRyner:
@ikatson 
![CleanShot 2024-08-29 at 17 56 42@2x](https://github.com/user-attachments/assets/e41af065-db8d-49f3-a415-7934d10dfc68)

![CleanShot 2024-08-29 at 17 56 28@2x](https://github.com/user-attachments/assets/61784f5a-ca26-4722-8524-7cefb15a996f)

Hm, it seems not to work in GUI

ikatson:
You're right! Thanks for spotting. Fixing in https://github.com/ikatson/rqbit/commit/6ca35612309c0608d8df6239b9c946692b3dc389, will merge into main once the tests finish https://github.com/ikatson/rqbit/actions/runs/10618548241

ikatson:
Merged into main

Pentaphon:
[...]

PR #223, opened 2024-08-28
[Feature] option to disable upload

ikatson:
context: https://github.com/ikatson/rqbit/issues/220

Usage:

```
rqbit --disable-upload ...
```

or environment variable

```
RQBIT_DISABLE_UPLOAD=true rqbit ...
```


DoctorRyner:
I just ran the same torrent multiple times with and without this option and it seems to work, great!

The only thing left is to add checkbox or something to Desktop GUI for this

PR #229, opened 2024-08-29
Standardized m3u8 formatting

Artrix9095:
Currently you're breaking the m3u8 standard and it breaks some media players. Adding the `#EXTM3U` tag should fix it. 

There is more information that *would* be best practice to add but I don't know how to add it. Please read the [m3u8 spec](https://www.rfc-editor.org/rfc/rfc8216) or the [wiki](https://en.wikipedia.org/wiki/M3U) for more information.

### DIFF
--- a/crates/librqbit/src/http_api.rs
+++ b/crates/librqbit/src/http_api.rs
@@ -190,6 +190,7 @@ impl HttpApi {
             let body = it
                 .into_iter()
                 .map(|(torrent_idx, file_idx, filename)| {
+                    // TODO: add #EXTINF:{duration} and maybe codecs ?
                     format!("http://{host}/torrents/{torrent_idx}/stream/{file_idx}/{filename}")
                 })
                 .join("\r\n");
@@ -201,7 +202,7 @@ impl HttpApi {
                         "attachment; filename=\"rqbit-playlist.m3u8\"",
                     ),
                 ],
-                body,
+                format!("#EXTM3U\r\n{body}"), // https://en.wikipedia.org/wiki/M3U
             )
         }
 


ikatson (approved):

Artrix9095:
Is there any way I can get the file size for each torrent?

ikatson:
What do you mean? Through http API? Must be in one of the stats endpoints, /torrents/id/stats/v1 if I remember correctly (on phone)

Artrix9095:
the `#EXTINF:` tag **requires** the file size for each uri in the playlist.

Artrix9095:
See https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.1

ikatson:
Build playlist content is called from the iterator iter_filenames_and_lengths, the second item there is length in bytes

ikatson:
> See https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.1

That's not what it is though, it's not HTTP live streaming streaming, just a playlist with links .

See https://en.m.wikipedia.org/wiki/M3U

So at best you can add #EXTBYT

There's no length in seconds, rqbit doesn't understand video/audio formats and their details. 

Streaming is done through range header (bytes) and a lot of clients are fine with this. 

So not sure what is it breaking, it's just probably not what you think it is?

Artrix9095:
My bad! I got the file size and video length mixed up. Currently it still breaks [Hls.js](https://github.com/video-dev/hls.js) which is a fairly popular HLS player on the web.

ikatson:
Got it. It's not HLS though which afaik needs specially crafted input, full files aren't enough

Artrix9095:
It is possible to split each file into segments relatively easily. Segments consist of .ts files which afaik are quite literally just chunked mpeg files (would be basically the same as what you're already doing just with a different file extension). I'll try to implement this in another PR if I can get some help

Artrix9095:
Actually this is might be more complicated and require FFmpeg. Which might add extra bloat. Probably should make this a feature flag called "pure-hls" or something

PR #230, opened 2024-09-01
Remove error when adding duplicate torrents

Artrix9095:
Error is kinda unnecessary here. Makes the user have to make another API request beforehand to check if the torrent is already managed. This can easily be simplified by just returning the managed torrent instead of an error.


ikatson:
I don't have a strong opinion on this, so if this caused trouble, no problem merging.

However there's a chance the torrent was added when different parameters,  e.g. output_folder.

So maybe at least return a different status code to indicate the difference between created and already exists

PR #233, opened 2024-09-08
In PEX we should also look at dropped peers

izderadicka:
As per BEP 11 - dropped peers are peers that have been active, but  not needed by remote peer anymore.  So it's good chance they will have a free slot, so we should try them too - before added peers


ikatson:
LGTM

PR #239, opened 2024-09-18
Change to restarting of dead peers

izderadicka:
For a while I was observing how dead peers are retried and found couple of thing, which I think are sub-optimal:
- when incoming peer has error, then `on_peer_died` is called and eventually peer is scheduled as outgoing,   The `SocketAddress` in this case  has outgoing port of peer I think, so retry with it as outgoing does not make sense to me.
- scheduled retries  stay pending even after the download of torrent is finished ( often with long waiting times). They keep popping out long after then.  I think it would be better to cancel them after download finishes.

This not final PR - rather a sketch to  illustrate my findings.  There is probably better way to solve it.

### DIFF
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -113,7 +113,7 @@ use super::{
 
 #[derive(Debug)]
 struct InflightPiece {
-    peer: PeerHandle, 
+    peer: PeerHandle,
     started: Instant,
 }
 
@@ -1001,7 +1001,7 @@ impl PeerHandler {
             }
         };
 
-        let error = match error {
+        let _error = match error {
             Some(e) => e,
             None => {
                 trace!("peer died without errors, not re-queueing");
@@ -1022,6 +1022,7 @@ impl PeerHandler {
 
         if self.incoming {
             // do not retry incoming peers
+            debug!("incoming peer {handle} died, not re-queueing");
             return Ok(());
         }
 
@@ -1031,16 +1032,11 @@ impl PeerHandler {
         drop(pe);
 
         if let Some(dur) = backoff {
-            debug!(
-                "{} peer {} died -  {error} and will retry in {dur:?}",
-                if self.incoming { "incoming" } else { "outgoing" },
-                self.addr
-            );
             self.state.clone().spawn(
                 error_span!(
                     parent: self.state.torrent.span.clone(),
                     "wait_for_peer",
-                    peer = handle.to_string(), 
+                    peer = handle.to_string(),
                     duration = format!("{dur:?}")
                 ),
                 async move {
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -438,11 +438,11 @@ impl TorrentStateLive {
         match res {
             // We disconnected the peer ourselves as we don't need it
             Ok(()) => {
-                handler.on_peer_died(None)?;
+                handler.on_peer_died(None, true)?;
             }
             Err(e) => {
                 debug!("error managing peer: {:#}", e);
-                handler.on_peer_died(Some(e))?;
+                handler.on_peer_died(Some(e), true)?;
             }
         };
         drop(permit);
@@ -505,11 +505,11 @@ impl TorrentStateLive {
         match res {
             // We disconnected the peer ourselves as we don't need it
             Ok(()) => {
-                handler.on_peer_died(None)?;
+                handler.on_peer_died(None, false)?;
             }
             Err(e) => {
                 debug!("error managing peer: {:#}", e);
-                handler.on_peer_died(Some(e))?;
+                handler.on_peer_died(Some(e), false)?;
             }
         }
         drop(permit);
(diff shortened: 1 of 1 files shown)


crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1013,22 +1013,37 @@ impl PeerHandler {
         if self.state.is_finished_and_no_active_streams() {
             debug!("torrent finished, not re-queueing");
             pe.value_mut().state.set(PeerState::NotNeeded, &pstats);
+            // also cancel all retried dead peers
+            if ! self.state.cancellation_token.is_cancelled() {
```
ikatson, line 1017: a few problems with this:
- this will cancel all tasks, not just peers. This is major and needs to be fixed
- you don't need to check is_cancelled() to cancel.
- I'd rather not use the cancellation token for this as it's specifically for destroying everything + it's a one-time thing, it's not reusable, i.e. you can't reinstate it after canceling. But a torrent being finished is not "the end". The user might change "only_files" for example.

In general I think these delayed reconnect attempts aren't harmful, especially when it will be done only once, and it's easier to keep it like it is

ikatson (changes requested):
Put some comments. Also recently the e2e test was very stable, but it broke in this PR, which makes me think it has a bug

crates/librqbit/src/torrent_state/live/mod.rs
izderadicka: I see,  
it was rather a test - to see if  scheduled retries are canceled - and  yes they are.
But as you said - much more is canceled too - for instance incoming peers are stop being served. Is actually same as pausing torrent.

Delayed reconnect are no problem as you stated,  I was just curious why they are appearing - even several hours after torrent is finished. It's rather small annoyance in log.  

Maybe to create a child cancellation token just for this purpose?

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -438,11 +438,11 @@ impl TorrentStateLive {
         match res {
             // We disconnected the peer ourselves as we don't need it
             Ok(()) => {
-                handler.on_peer_died(None)?;
+                handler.on_peer_died(None, true)?;
```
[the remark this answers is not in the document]
izderadicka: I'll look into it.

izderadicka:
If you don't mind, I'll play more around this PR - remove first cancellation as it it  clearly wrong now and look only on preventing retry on failed incoming peers.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1018,17 +1020,27 @@ impl PeerHandler {
 
         pe.value_mut().state.set(PeerState::Dead, &pstats);
 
+        if self.incoming {
```
ikatson, line 1023: please put a debug!() line here

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1018,17 +1020,27 @@ impl PeerHandler {
...
+            return Ok(());
+        }
+
         let backoff = pe.value_mut().stats.backoff.next_backoff();
 
         // Prevent deadlocks.
         drop(pe);
 
         if let Some(dur) = backoff {
+            debug!(
```
ikatson, line 1034: please remove this message entirely, it's just noise.

The error_span!() below does it all already, and it's printed multiple times as part of spawn().

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -999,7 +1001,7 @@ impl PeerHandler {
             }
         };
 
-        let _error = match error {
+        let error = match error {
```
ikatson, line 1004: why? It will complain about unused local variable. Please revert.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -113,7 +113,7 @@ use super::{
 
 #[derive(Debug)]
 struct InflightPiece {
-    peer: PeerHandle,
+    peer: PeerHandle,
```
ikatson, line 116: formatting - space at the end is junk

ikatson (changes requested):

crates/librqbit/src/torrent_state/live/mod.rs
ikatson: Looking at the surprising test failure - you might need to call self.state.peers.drop_peer(handle) before returning. Not sure, please take a look.

izderadicka: Test now works, though I do not understand the reason for failure - when only incoming peers were retried - I just then made some changes to logging and now cannot recreate test failure.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1018,6 +1020,12 @@ impl PeerHandler {
 
         pe.value_mut().state.set(PeerState::Dead, &pstats);
 
+        if self.incoming {
+            // do not retry incoming peers
+            debug!("incoming peer {handle} died, not re-queueing");
```
ikatson, line 1025: nit: at least for the sake of consistent style please put handle into a parameter, e.g. debug!(peer=handle, "incoming peer died, not re-queueing");

ikatson (approved):

PR #241, opened 2024-09-20
wip: add explanations about work to do

cocool97:
This PR is a work in progress to allow storage to delete pieces when they have been served in streaming context


cocool97:
@ikatson Can you confirm if expected changes seem good to you ? I'll implement it afterwards

ikatson:
I won't feel comfortable merging this. As I think I mentioned previously, streaming with garbage collection is a major design change and can't be easily hacked into the codebase.

For example, given your comment in this PR - it's not good for separation of concerns. Streaming.rs shouldn't be deleting anything or knowing anything about piece garbage collection.

ikatson:
Also:

> on_piece_completed() could thus be replaced by delete_chunk()

Same comment here. librqbit shouldn't be deleting chunks. At most you could provide enough integration points so that your custom Box<dyn Something> implementation (e.g. Box<dyn TorrentStorage>, although that won't be enough) can do deletion without the core knowing anything about it, but also not breaking.

ikatson:
One way to try to make this work is:
- add "streaming_only: bool" to the torrent.
- in all the "stream then delete" places check for "streaming_only", and branch accordingly.

At least this would make clear what needs to be changed and where. I'm not saying I'll merge it if it's done this way, but at least will give more room for discussion. Piece-meal tiny PRs aren't enough to show the full picture of how it's all going to work in the end.

Writing up a design doc upfront would also help not waste time.

PR #242, opened 2024-09-22
fix: log levels not coherent

cocool97:
In `dht` crate some log levels do not reflect what they are indeed saying. This is quite annoying because some logs are in error when being really debug logs, and this floods logging and thus prevent from filtering nicely with crates like `env_logger`.

This PR improves that

### DIFF
--- a/crates/dht/src/dht.rs
+++ b/crates/dht/src/dht.rs
@@ -42,7 +42,7 @@ use tokio::{
 };
 
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, debug_span, error, error_span, info, trace, warn, Instrument};
+use tracing::{debug, debug_span, error, error_span, info, trace, trace_span, warn, Instrument};
 
 #[derive(Debug, Serialize)]
 pub struct DhtStats {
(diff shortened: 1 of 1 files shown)


ikatson:
tracing crate is confusing regarding the levels - as it has both a span level and event level (unlike regular logging libraries which only have even level).

Things were done they way they are on purpose, and it seems this PR makes it worse in both dimensions:
1. error_span!() is everywhere, so that the span and its fields are visible in all log lines. This is to create a hierarchy of log lines, so that you can see what was spawned by what.
2. non-critical errors like DHT errors are all on debug!() level so that they don't noise the main "rqbit server start" output which is at INFO level.

So your change: 
- adds noise by increasing the verbosity of DHT considerably (in default INFO level), which starts printing garbage to stdout when the user doensn't care
- even when it does print these noisy lines, your change removes useful information from those lines' spans.

All of which makes me confused as to why you claim it's better. Can you elaborate?

Maybe it all works differently with non-tracing crates like env_logger?

cocool97:
env_logger only filters at event level, and all the lines I changed produce logs that are at the wrong level. 

For example, errors are printed as DEBUG level, so that they are not displayed when filtering at INFO. Same for "find_node" logs, that are currently logged using error_span! macro when these are definitely not errors.

ikatson:
Ok, I now see what the problem is after trying "env_logger" create with "tracing = {features = ["log"]}" dependency.

Spans get logged (not just events), and produce noise in that case. It doesn't happen when using tracing_subscriber

I suggest you just use tracing_subscriber::fmt::init() which will work much better if you want librqbit to output logs. Won't be merging this.

cocool97:
😅
This will make everyone wanting to use this crate to also import `tracing_subcriber` crate ! Definitely don't think this is a good idea but so on

ikatson:
Then you need to find a way to fix it for your use-case without degrading the main rqbit usage. So far your PR makes degrades it along 2 axises that I mentioned

PR #245, opened 2024-09-28
Adds option to emit json structured logs to stdout

scottopell:
This adds a flag `--structured-stdout` that will add a layer emitting to stdout with json-structure. This is useful in environments where logs are collected and ingested to more easily parse and query.

This filter does respect the `log_file_rust_log` option, so I am open to renaming that option to be more agnostic, but it seems fine as-is to me.


crates/librqbit/src/tracing_subscriber_config_utils.rs
```diff
@@ -52,7 +52,8 @@ impl std::io::Write for Writer {
 pub struct InitLoggingOptions<'a> {
     pub default_rust_log_value: Option<&'a str>,
     pub log_file: Option<&'a str>,
-    pub log_file_rust_log: Option<&'a str>,
```
ikatson, line 55: Why change this? It broke the desktop/ app.

crates/librqbit/src/tracing_subscriber_config_utils.rs
```diff
@@ -107,7 +108,25 @@ pub fn init_logging(opts: InitLoggingOptions) -> anyhow::Result<InitLoggingResul
...
                     .with_filter(
                         EnvFilter::builder()
-                            .parse(opts.log_file_rust_log.unwrap_or("info,librqbit=debug"))
+                            .parse(opts.log_file_rust_log)
+                            .context("can't parse log-file-rust-log")?,
+                    ),
+            )
+            .try_init()
+            .context("can't init logging")?;
+    } else if opts.structured_stdout {
```
ikatson, line 117: Why is it either logfile or structured stdout? Why not support both?

ikatson (changes requested):
Needs a fix, at least to unbreak the desktop app.

crates/librqbit/src/tracing_subscriber_config_utils.rs
ikatson: If this is because of typing looks like you can use [```Option<Layer>```](https://stackoverflow.com/questions/77607599/how-to-add-a-layer-to-tracing-subscribers-registry-based-on-a-condition)

scottopell:
Thanks for the feedback and pointers, I'm moving this back to a draft for now while I fight with tracing_subscriber

ikatson:
Also please don't change stderr to stdout for logs. If you really need it for some reason, make it a parameter, but keep the current behavior.

scottopell:
I'm misunderstanding where the stdout log stream(s) are coming from, I'm seeing duplicated logs, some with human and some with json. I guess need to revisit all the branches of the logging recorder and see what I missed
```
$ cargo run -- --log-format=json server start /tmp/scratch 2&>/dev/null
2024-10-03T15:55:58.558134Z  INFO rqbit: inreased open file limit limit=61440
{"timestamp":"2024-10-03T15:55:58.558237Z","level":"INFO","fields":{"message":"inreased open file limit","limit":61440},"target":"rqbit"}
```

crates/librqbit/src/tracing_subscriber_config_utils.rs
```diff
@@ -112,6 +122,7 @@ pub fn init_logging(opts: InitLoggingOptions) -> anyhow::Result<InitLoggingResul
                 fmt::layer()
                     .with_ansi(false)
                     .with_writer(log_file)
+                    .and_then(if let Format::Json = opts.format { Some(fmt::layer().json())} else { None } )
```
ikatson, line 125: Not sure if it makes sense to change the log file to JSON at the same time as stdout.
One way to do this is 2 separate parameters:

--log-json (for stderr)
--log-file-log-json (for file log to be json)

crates/librqbit/src/tracing_subscriber_config_utils.rs
```diff
@@ -81,7 +87,11 @@ pub fn init_logging(opts: InitLoggingOptions) -> anyhow::Result<InitLoggingResul
 
     let layered = tracing_subscriber::registry()
         // Stderr logging layer.
-        .with(fmt::layer().with_filter(stderr_filter))
+        .with(
+            fmt::layer()
+                .with_filter(stderr_filter)
+                .and_then(if let Format::Json = opts.format { Some(fmt::layer().json())} else { None } ),
```
ikatson, line 93: I'm not sure and_then works the way you think it does. You are probably logging to both layers - the JSON one and the regular one.

You probably need to make stderr and JSON layers mutually exclusive like [here](https://github.com/tokio-rs/tracing/issues/894):

```
let (json, plain) = if enable_json {
    (Some(tracing_subscriber::fmt::layer().json()), None)
} else {
    (None, Some(tracing_subscriber::fmt::layer()))
};
```

ikatson (changes requested):

crates/librqbit/src/tracing_subscriber_config_utils.rs
```diff
@@ -107,35 +114,43 @@ pub fn init_logging(opts: InitLoggingOptions) -> anyhow::Result<InitLoggingResul
...
         ));
-        layered
-            .with(
+        let log_env_filter = EnvFilter::builder()
+            .parse(opts.log_file_rust_log.unwrap_or("info,librqbit=debug"))
+            .context("can't parse log-file-rust-log")?;
+        let log_layer: Box<dyn Layer<_> + Send + Sync> = if opts.log_file_json {
+            Box::new(
                 fmt::layer()
-                    .with_ansi(false)
```
ikatson, line 113: this makes log file contain ANSI escape codes which makes it hard to grep and use. Those escape codes are intended for terminal use only.

PR LGTM otherwise, I'd merge if not for this

ikatson (changes requested):

PR #250, opened 2024-10-03
refactor: use byteorder crate to enhance portability (to BE systems), and enhance code readability, avoiding manual byte manipulations

PastaPastaPasta:
Note: this also fixes what I think is a small bug here: https://github.com/ikatson/rqbit/pull/250/files#diff-b4f79704f8a70bb6832cd8eb994612ba62b1e199efcebd39228d0761a257920dR285

where `6` is always printed instead of the invalid size like other instances do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2596,6 +2596,7 @@ version = "5.1.0"
 dependencies = [
  "anyhow",
  "backoff",
+ "byteorder",
  "bytes",
  "chrono",
  "dashmap 6.1.0",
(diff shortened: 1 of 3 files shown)


ikatson:
nit: if I wrote this today, I'd just use std's to_be_bytes / from_be_bytes to avoid adding a dependency. I guess I didn't know about these functions when I wrote this originally years ago. 
Also afaik your change doesn't increase portability as the commit name states, it should have worked on be systems already (in theory).

But the change looks ok anyway, the above is just FYI, merging

PR #254, opened 2024-10-12
Fix: transactional pause

Mrreadiness:
Hi! Thanks for the great library!
I've recently faced a bug when pausing while initializing torrent returned an error, but it changed the internal flag. This fix should prevent such behavior.

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -1365,16 +1365,17 @@ impl Session {
     }
 
     pub async fn pause(&self, handle: &ManagedTorrentHandle) -> anyhow::Result<()> {
-        let mut g = handle.locked.write();
-        let prev = g.paused;
-        g.paused = true;
-        drop(g);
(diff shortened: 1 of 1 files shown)


ikatson (approved):

crates/librqbit/src/session.rs
```diff
@@ -1365,8 +1365,20 @@ impl Session {
     }
 
     pub async fn pause(&self, handle: &ManagedTorrentHandle) -> anyhow::Result<()> {
-        handle.locked.write().paused = true;
-        handle.pause()?;
+        let prev_state;
```
ikatson, line 1368: Can it be simplified by first calling handle.pause(), and then only if it was successful set paused = true?

This way the lock is taken only once

Mrreadiness: Yeah, it makes sense. Updated.

ikatson (approved):

PR #261, opened 2024-10-23
PEX - sharing peers with othes

izderadicka:
For PEX only receiving part was implemented,  this is an attempt also to send PEX extended message to connected peers.

It's WIP now,  there is a problem with E2E test , which is unstable now,  often ends with :
```
---- tests::e2e::test_e2e_download stdout ----
[crates/librqbit/src/tests/e2e.rs:65:9] tempdir.path() = "/tmp/rqbit_e2eYmBFKD"
thread 'tests::e2e::test_e2e_download' panicked at crates/librqbit/src/tests/test_util.rs:213:6:
called `Result::unwrap()` on an `Err` value: wait_until timeout: last result = Some(metrics.num_alive_tasks() = 0, expected 1)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

which does not make sense to me -   how could be number of live tasks in runtime 0?   

But there is also another interesting behavior of e2e test if I change **initial delay on PEX send to 0** (here https://github.com/izderadicka/rqbit/blob/82914bac27226b1d10157832ee66aeb7af67ca18/crates/librqbit/src/torrent_state/live/mod.rs#L807), then:

1) Sometime some server tasks did not end (5 tasks per server) - as if server session was not dropped - it's kind of arbitrary - message same as above but now number of task is 5 or 10, or n x 5 - depends (max I seen was 80). 
2) Servers send PEX msg between themselves - which is strange - as their torrent is fully downloaded - so they should not open new connections and peers received from client should just be stored as `not_needed` , but should not initiate  connection try, at least this is how I understand the code.

### DIFF
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -803,7 +803,17 @@ impl TorrentStateLive {
     ) -> anyhow::Result<()> {
         let mut sent_peers_live: HashSet<SocketAddr> = HashSet::new();
         const MAX_SENT_PEERS: usize = 50; // As per BEP 11 we should not send more than 50 peers at once (here it also applies to fist message, should be OK as we anyhow really have more)
+        const PEX_MESSAGE_INTERVAL: Duration = Duration::from_secs(60); // As per BEP 11 recommended interval is min 60 seconds
+        let mut delay = Duration::from_secs(10); // Wait 10 seconds before sending the first message to assure that peer will stay with us
+
         loop {
+            tokio::select! {
+                _ = tx.closed() => return Ok(()),
+                _ = tokio::time::sleep(delay) => {},
+
+            }
+            delay = PEX_MESSAGE_INTERVAL;
+
             let addrs_live_to_sent = self
                 .peers
                 .states
@@ -813,7 +823,16 @@ impl TorrentStateLive {
                     let addr = peer.outgoing_address.as_ref().unwrap_or_else(|| e.key());
 
                     if *addr != peer_addr {
-                        if peer.state.is_live() && !sent_peers_live.contains(addr) {
+                        let has_outgoing_connections = peer
+                            .stats
+                            .counters
+                            .outgoing_connections
+                            .load(Ordering::Relaxed)
+                            > 0; // As per BEP 11 share only those we were able to connect
+                        if peer.state.is_live()
+                            && has_outgoing_connections
+                            && !sent_peers_live.contains(addr)
+                        {
                             Some(*addr)
                         } else {
                             None
@@ -822,7 +841,7 @@ impl TorrentStateLive {
                         None
                     }
                 })
-                .take(50) 
+                .take(50)
                 .collect::<HashSet<_>>();
 
             let addrs_closed_to_sent = sent_peers_live
@@ -838,28 +857,21 @@ impl TorrentStateLive {
                 .take(MAX_SENT_PEERS)
                 .collect::<HashSet<_>>();
 
-            // BEP 11 - Dont send closed if they are now in live 
-            // it's assured by mutual exclusion of two  above sets  if in sent_peers_live, it cannot be in addrs_live_to_sent, 
+            // BEP 11 - Dont send closed if they are now in live
+            // it's assured by mutual exclusion of two  above sets  if in sent_peers_live, it cannot be in addrs_live_to_sent,
             // and addrs_closed_to_sent are only filtered addresses from sent_peers_live
 
             if !addrs_live_to_sent.is_empty() || !addrs_closed_to_sent.is_empty() {
-                let pex_msg = extended::ut_pex::UtPex::from_addrs(&addrs_live_to_sent, &addrs_closed_to_sent);
+                debug!(peer=?peer_addr, "sending PEX with {} live ({:?})and {} closed peers", addrs_live_to_sent.len(), addrs_live_to_sent,addrs_closed_to_sent.len());
+                let pex_msg =
+                    extended::ut_pex::UtPex::from_addrs(&addrs_live_to_sent, &addrs_closed_to_sent);
                 let ext_msg = extended::ExtendedMessage::UtPex(pex_msg);
                 let msg = Message::Extended(ext_msg);
 
-                tx.send(WriterRequest::Message(msg))?;
-
-                debug!(peer=?peer_addr, "sending PEX with {} live and {} closed peers", addrs_live_to_sent.len(), addrs_closed_to_sent.len());
-                sent_peers_live.extend(&addrs_live_to_sent);
-                sent_peers_live.retain(|addr| !addrs_closed_to_sent.contains(addr));
-
-
-            }
-
-            tokio::select! {
-                _ = tx.closed() => return Ok(()),
-                _ = tokio::time::sleep(Duration::from_secs(60)) => {},
-
+                if tx.send(WriterRequest::Message(msg)).is_ok() {
+                    sent_peers_live.extend(&addrs_live_to_sent);
+                    sent_peers_live.retain(|addr| !addrs_closed_to_sent.contains(addr));
+                }
             }
         }
     }
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -841,7 +841,7 @@ impl TorrentStateLive {
                         None
                     }
                 })
-                .take(50)
+                .take(MAX_SENT_PEERS)
                 .collect::<HashSet<_>>();
 
             let addrs_closed_to_sent = sent_peers_live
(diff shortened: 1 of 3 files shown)


izderadicka:
Actually I see same problems also on `main` branch - can you try on main

```shell
#!/usr/bin/env bash

# end on sigint
STOP=""
trap "STOP=1" INT


NUM_RUNS=${1:-100}
TOTAL=$NUM_RUNS
FAILURES=0
export NO_COLOR=1
rm /tmp/librqbit_e2e_failures.log
while [ $NUM_RUNS -gt 0 ]; do
    echo Running test $((TOTAL - NUM_RUNS + 1)) of $TOTAL
    NUM_RUNS=$((NUM_RUNS - 1))
    cargo test --package librqbit  tests::e2e::test_e2e_download > /tmp/librqbit_e2e_run.log
    res=$?
    if [ $res -ne 0 ]; then
        FAILURES=$((FAILURES + 1)) 
        echo "FAILURE with code $res" >> /tmp/librqbit_e2e_failures.log
        tail -50 /tmp/librqbit_e2e_run.log >> /tmp/librqbit_e2e_failures.log
        echo "----------------------------------------------" >> /tmp/librqbit_e2e_failures.log

    fi
    if [ "$STOP" = "1" ]; then
        break
    fi
done
rm /tmp/librqbit_e2e_run.log
echo "Total runs : $TOTAL, failures : $FAILURES"
echo "Failures log /tmp/librqbit_e2e_failures.log" 
```

On main branch I got 28 failures out of 100 - mainly above mentioned 0 tasks,  but couple of dropcheck failures.

ikatson:
Maybe some tokio upgrade broke it, I'll take a look someday. Just got back from a 10 day break, so catching up.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -795,6 +795,86 @@ impl TorrentStateLive {
...
+                })
+                .take(50)
+                .collect::<HashSet<_>>();
+
+            let addrs_closed_to_sent = sent_peers_live
+                .iter()
+                .filter(|addr| {
+                    self.peers
+                        .states
+                        .get(addr)
```
ikatson, line 852: you're locking the state twice, and it may race - e.g. it'll be live in first pass where you are computing "addrs_live_to_sent" and then not live later, or the other way around.

It's hard to reason about what will happen in this case, esp. when below you're trying to provide strong guarantees like "it's assured by mutual exclusion of two  above sets".

So overall, how about refactor this for simplicity:

1. store live peer's view in "sent_peers_live" as you are doing now.
1. for each loop iteration
    1. collect **all** live addrs into HashSet (not bounded by 50, but bounded by the semaphore anyway, 128 if I remember correctly). Let's name it "all_live"
    1. connected = (all - sent).take(50)
    2. dropped = (sent - all).take(50)
    3. sent = sent + connected - dropped

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -1717,6 +1806,7 @@ impl PeerHandler {
         B: AsRef<[u8]> + std::fmt::Debug,
     {
         // TODO: this is just first attempt at pex - will need more sophistication on adding peers - BEP 40,  check number of live, seen peers ...
+        debug!("received PEX message with {} added peers and {} dropped peers", msg.added_peers().count(), msg.dropped_peers().count());
```
ikatson, line 1809: every message is already logged, you don't need this?

ikatson (changes requested):

ikatson:
As for flaky tests, it seems stable now on my box (without extensive testing). The fixes are here:
https://github.com/ikatson/rqbit/commit/e371522a8014c6739209b29a88cb36e2686c8d20
https://github.com/ikatson/rqbit/commit/87e09a60f7b9d457ca306cc30b24c4d1f0ac7a17

Can you check if it's still bad for u?

crates/librqbit/src/torrent_state/live/mod.rs
izderadicka: It was just for debugging - to understand what other peers are sending.
Interestingly some peers were sending completely empty PEX message with no added nor dropped peers, which is against spec.
Removed.

izderadicka:
> As for flaky tests, it seems stable now on my box (without extensive testing). The fixes are here: [e371522](https://github.com/ikatson/rqbit/commit/e371522a8014c6739209b29a88cb36e2686c8d20) [87e09a6](https://github.com/ikatson/rqbit/commit/87e09a60f7b9d457ca306cc30b24c4d1f0ac7a17)
> 
> Can you check if it's still bad for u?

Checked on latest main and no problem,   only in one run test complained about `Blocking waiting for file lock on package cache` 3 times,  but test succeed then - but at that time my notebook was heavily loaded but these tests and some other work, so I guess that was the reason.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -795,6 +795,86 @@ impl TorrentStateLive {
...
+                        {
+                            Some(*addr)
+                        } else {
+                            None
+                        }
+                    } else {
+                        None
+                    }
+                })
+                .take(50)
```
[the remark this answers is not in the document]
izderadicka: There is also condition `&& !sent_peers_live.contains(addr)` on line 834,  which filters out already sent addresses -  so what I got are next 50, which **have not been sent to peer** - I looked on it in swarm on busy torrent (Ubuntu) and it seemed do work.
Anyhow this is also related to the other comment, so it may change.

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -795,6 +795,86 @@ impl TorrentStateLive {
...
+            // and addrs_closed_to_sent are only filtered addresses from sent_peers_live
+
+            if !addrs_live_to_sent.is_empty() || !addrs_closed_to_sent.is_empty() {
+                debug!("sending PEX with {} live ({:?})and {} closed peers", addrs_live_to_sent.len(), addrs_live_to_sent,addrs_closed_to_sent.len());
+                let pex_msg =
+                    extended::ut_pex::UtPex::from_addrs(&addrs_live_to_sent, &addrs_closed_to_sent);
+                let ext_msg = extended::ExtendedMessage::UtPex(pex_msg);
+                let msg = Message::Extended(ext_msg);
+
+                if tx.send(WriterRequest::Message(msg)).is_ok() {
```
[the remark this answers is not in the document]
izderadicka: Could be - It'll leave anyhow on line 811 as here we check `tx.closed()`,  leaving on error will make it more explicit, so I change it (it took me while to understand why I did not do it initially, so it's not probably good code, as it's not obvious).

crates/librqbit/src/torrent_state/live/mod.rs
```diff
@@ -795,6 +795,86 @@ impl TorrentStateLive {
...
+
+    async fn task_send_pex_to_peer(
+        self: Arc<Self>,
+        peer_addr: SocketAddr,
+        tx: PeerTx,
+    ) -> anyhow::Result<()> {
+        let mut sent_peers_live: HashSet<SocketAddr> = HashSet::new();
+        const MAX_SENT_PEERS: usize = 50; // As per BEP 11 we should not send more than 50 peers at once (here it also applies to fist message, should be OK as we anyhow really have more)
+        const PEX_MESSAGE_INTERVAL: Duration = Duration::from_secs(60); // As per BEP 11 recommended interval is min 60 seconds
+        let mut delay = Duration::from_secs(10); // Wait 10 seconds before sending the first message to assure that peer will stay with us
```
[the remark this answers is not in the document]
izderadicka: Main reasons for using `sleep`:
- initial wait -  do not want to start immediately after peer connection,  as peer can quickly disconnect.  Only want to send PEX to 'stable' peers.  interval will fire immediately.
- changing wait period -  now it's 10, then 60 secs.   But can be extended - I was thinking about some extension of the interval - up to 600 secs or something like this.

From this perspective sleep seems easier to use.

crates/librqbit/src/torrent_state/live/mod.rs
izderadicka: I see - but I think it's not the case here because liveness check actually happens only once, depending on content of `sent_peers_live` - which reflects already sent peers addresses:
- if address is in `sent_peers_live`, then its effectively checked for liveness only on line 853 `.map(|p| !p.value().state.is_live())`  (for closed peers), because of line 834 `&& !sent_peers_live.contains(addr)` which excludes it ultimately from `addrs_live_to_sent` set.
- if address is not in `sent_peers_live` -  it's checked only for `addrs_live_to_sent` on line 832 `if peer.state.is_live()` and second check for closed peers is not happening at all, because it's in iteration over `sent_peers_live`.

So in any case in each loop run **only one liveness check happens**, which is relevant for logic.

ikatson: My main point was that the logic is hard to read, and I proposed a simpler way to write it. All the state that you need to keep in your mind to understand why the liveness check is happening only once might serve as proof of that :)

izderadicka: I do not think that logic is difficult -   it's just all about `sent_peers_live`, which we need anyhow - if address is not in it, we check if we can add new address, which is live and not in it,   if address is in it, we check if we should remove if it's not live. That's it.  

Proposed change will work exactly same,   I think. Only  we will need always to collect all live peers, which is not "needed", we just need those, that are not in `sent_peers_live`.

izderadicka:
Closing as work continues in #284

PR #274, opened 2024-11-09
Script to run e2e test in loop to detect flaky test

izderadicka:
I used this script to test if e2e test is not flaky - thought that it might be useful to have it around for future.

### DIFF
--- a/scripts/test-e2e.sh
+++ b/scripts/test-e2e.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# end on sigint
+STOP=""
+trap "STOP=1" INT
+
+
+NUM_RUNS=${1:-100}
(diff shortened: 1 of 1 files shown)


ikatson:
This is just personal taste, but this script is a bit too complex for my liking. I'd rather not merge it in at all. At most a one-liner that we can paste into Makefile maybe?

I remember previously I showed a one-liner that I was using instead of a complex bash script that did the same job, wdyt of adding that to Makefile (to quickly run smth like ```make e2e_until_fail``` instead?

izderadicka:
OK -  here are couple things, that makes it bit more complex -   defined number of runs and measuring failure rate and storing log from all failures -   especially last one was useful as I was able to see different failure modes.

izderadicka:
Closing as not fitting

PR #283, opened 2024-11-24
Enable thin LTO release but not release-github

luigi311:
Enable lto thin mode to reduce sizes
release
none -> thin
size: 217.7 mb -> 186.5 mb
time: 50s -> 51s

thin -> fat
size 186.5 mb -> 150.5 mb
time 51s -> 2m 

Looks like release-github doesnt require it though as thin lto actually increases the binary size with debug disabled

release-github
none -> thin
size: 15.3 mb -> 16.4 mb
time: 37s -> 38s 

thin -> fat
size: 16.4 mb -> 12.8
time: 38s -> 1m 15s

tauri build release mode also has a pretty big drop
305.3 mb -> 237.8 mb

tauri build with debug false
18.5 mb -> 20.1 mb

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,4 @@ lto = "thin"
 [profile.release-github]
 inherits = "release"
 debug = false
+lto = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ default-members = [
 [profile.release]
 panic = "abort"
 debug = true
+lto = "thin"
 
 [profile.release-github]
 inherits = "release"


luigi311:
While lto didn't reduce the size of release-github I wonder what the performance looks like if any difference at all.

ikatson:
Thank you so much for the sponsorship, it really made my day! I'm currently traveling until Friday and may have limited availability to review your PRs closely, but I'll do my best to take a look before then.

luigi311:
No worries theres no pressure feel free to take your time, none of my changes are anything major just figured i would take a look at some simpler stuff and see if its worth it or not.

ikatson:
How about build times? I'd argue this might be even more important than binary size.

Wdyt of setting them in reverse this:

- "release" to no lto (assuming that's the fastest build time). So that locally you can "cargo build --release" and it's not taking forever.
- "release-github" to lto, cause we don't care about release build times as much in CI. Probably thin, as a balance between build time, build size and final performance.

luigi311:
Build times are what measured for time in the original description sorry should of made that clear. Its the reason i left the default to thin instead of fat since that inflates the build times like crazy. 

I still need to take a look to see what performance looks like between the two to make sure this doesnt cause any regressions. Is there a method you like to use for performance testing? I see your cargo test are pretty extensive, should I use that as a reference?

ikatson:
> I still need to take a look to see what performance looks like between the two to make sure this doesnt cause any regressions. Is there a method you like to use for performance testing? I see your cargo test are pretty extensive, should I use that as a reference?

That's probably unnecessary, as there's no reference to compare against - there's no official "benchmark" for perf testing rqbit.

For larger / core changes I usually run "make devserver-profile", download something popular, then look at CPU and profiler output not to contain obvious hotspots.

PR #284, opened 2024-11-28
PEX - sharing peers with othes [continue #261]

ikatson:
#261 isn't being worked on, so I tried to finish it here

- [ ] Do not compute live peers within every peer all the time - store them in a shared place

CC @izderadicka

### DIFF
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -884,7 +884,7 @@ impl TorrentStateLive {
                         None
                     }
                 })
-                .take(50)
+                .take(MAX_SENT_PEERS)
                 .collect::<HashSet<_>>();
 
             let addrs_closed_to_sent = sent_peers_live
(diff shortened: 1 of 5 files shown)


izderadicka:
@ikatson Sorry, did not have much time recently -  I see that you rewrite the logic as per discussion in #261 

Logic is clear (though I'm still not convinced that previous logic was wrong).   I like the idea having shared view of live peers - so it'll have also some timestamp, where it was last updated and if it will be older that X secs then refreshed?   I guess it should be under RWLock to enable sharing and updates?

ikatson:
@izderadicka I didn't mean the previous logic was wrong, just that it was hard to read.

Yeah I don't want to merge it until the live peer computation is optimized: there may be tens of thousands of known peers, but only a limited number of live ones. It doesn't make sense to iterate tens of thousands to get a few live peers.

Yeah, RwLock is one option. Can also use a broadcast channel or smth. RwLock probably simpler

ikatson:
@izderadicka did that change, was harder than I thought, needed to refactor a few things here and there.

Seems to work perfectly.

Issue #286, opened 2024-12-02
Allow `ListOnly` request to time-out gracefully in web-ui when adding torrents


scottopell:
This is split off from https://github.com/ikatson/rqbit/issues/285#issuecomment-2509385742

When I try to add a magnet link for an old / low-seed count torrent, the current Upload modal requires a `ListOnly` request to complete before proceeding to the next step (file selection) of the Add Torrent workflow.
Currently, this request prevents proceeding to the next step of adding the torrent via the web interface, so certain `magnet` links cannot be added.

To address this, the `ListOnly` request should have a graceful timeout and the File Selection step should be skipped, just selecting an output destination.

scottopell:
Ah and I forgot that @ikatson also pointed out I'll need some changes in the backend to let the magnet resolution fail gracefully too

> you need to update the backend to allow adding torrents without waiting for them to resolve the magnet.
https://github.com/ikatson/rqbit/issues/285#issuecomment-2509393464

PR #287, opened 2024-12-02
Allows graceful timeout of listing files when adding a torrent via web-ui.

scottopell:
Fix for #286 

Current state:

- [x] Add a timeout for http requests
- [x] Use timeout in `UploadButton` and correct types
- [ ] File Selection modal should detect that no `ListTorrentResponse` is present and only present file destination picker

### DIFF
--- a/crates/librqbit/webui/src/api-types.ts
+++ b/crates/librqbit/webui/src/api-types.ts
@@ -114,6 +114,7 @@ export interface ErrorDetails {
   status?: number;
   statusText?: string;
   text: string | React.ReactNode;
+  timedOut: boolean;
 }
 
 export type Duration = number;
@@ -193,7 +194,8 @@ export interface RqbitAPI {
   ) => string | null;
   uploadTorrent: (
     data: string | File,
-    opts?: AddTorrentOptions,
+    opts: AddTorrentOptions,
+    timeout?: number | undefined
   ) => Promise<AddTorrentResponse>;
 
   pause: (index: number) => Promise<void>;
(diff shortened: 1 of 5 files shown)


ikatson:
Hmm, I think doing it in the backend is the only correct way. The backend has a "timeout" query parameter that you can pass, and then you don't need to make such a heavy change to UI code. This is the first thing I would do. Although I'm not sure what a reasonable default would be. I'm fine having NO timeout by default. So if you need it, maybe let the user configure it in UI when adding a torrent? E.g. the user clicks "add torrent through magnet", and in the dialog there's an option to set the timeout, which by default is empty.

But a better change would be to add a new way of adding torrent to the backend - this will spawn a tokio task to add a torrent with a larger timeout. The problem is that it won't show up in UI at all until resolved, and won't survive restarts.
For it to do both, we need a new torrent state, or maybe a separate store for torrents that are "being resolved".

ikatson:
> The backend has a "timeout" query parameter that you can pass, and then you don't need to make such a heavy change to UI code.

Uh, I remember writing this code, but I don't see it in main. I guess I never merged it in :D

ikatson:
Btw, if you add torrent without listing, it will still "block" the request until the magnet it resolved. And I guess if you close the browser window, it will cancel the request, which will cancel the future that was adding the torrent.

So the only reasonable way to do it is to spawn a task that would do it all. But then you'd need something in the backend to prevent spawning futures for adding the same torrents.

All this TBH smells like a moderately sized backend change that must be made first without touching the UI. So I'd pause for now and design this upfront in more detail before making the changes

ikatson:
> All this TBH smells like a moderately sized backend change that must be made first without touching the UI. So I'd pause for now and design this upfront in more detail before making the changes

FYI I'm trying to do this backend change and it's quite large and complex. Breaks a lot of assumptions and touches everything

PR #309, opened 2025-01-20
Initial support for interface binding

Quackdoc:
Currently a draft, this is "working" sort of when you pass the interface to the tracker, without passing the interface and binding to it, it announces the wrong thing. As such setting initial-peer will work, but no trackers will see the client.

I haven't really proof read any of the changes yet, this is just so I can get help with the main issue.

Additionally this PR makes changes to how tcp port binding works. if a user wishes to select a single port (IE. port forwarding) we can easily do this by passing the port..port + 1 due to how ranges work. the changes are purely convenience. Min port being greater then max port will give a range with nothing in it, so doing this will at least pass a single port, but perhaps throwing a warning or even an error would be better.

- [ ] don't clone so much
- [ ] figure out how pass interface to tracker

the second one I am really struggling with as it seems like it may require quite a bit of things to change.

### DIFF
--- a/crates/dht/src/dht.rs
+++ b/crates/dht/src/dht.rs
@@ -1138,6 +1138,7 @@ pub struct DhtConfig {
     pub bootstrap_addrs: Option<Vec<String>>,
     pub routing_table: Option<RoutingTable>,
     pub listen_addr: Option<SocketAddr>,
+    pub interface_name: Option<String>,
     pub peer_store: Option<PeerStore>,
     pub cancellation_token: Option<CancellationToken>,
 }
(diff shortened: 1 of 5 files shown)


ikatson:
"Don't clone as much" - I wouldn't worry about it before it all works

"Figure out how to pass interface to tracker" -  by a quick look there's nothing to "refactor", just extend. The only reasons to refactor would be growing code complexity but that can be addressed on top. 

There's 2 things: the "reqwest::Client" passed for HTTP trackers - that needs to be extended with "interface". The second is for UDP trackers need to pass down interface.

I would also resolve interface to index was early as possible and use that instead of name everywhere (if that works)

ikatson:
Fixed in https://github.com/ikatson/rqbit/pull/445

PR #330, opened 2025-02-20
Add support for P2P blocklists 

xobust:
This PR adds basic support for [P2P plaintext blacklists ](https://en.wikipedia.org/wiki/PeerGuardian#P2P_plaintext_format).
A list is downloaded at startup and stored in a interval tree. When an incoming connection is made the IP will be checked and connection aborted before handshake. 
Not sure how useful these lists are anymore but I wanted to learn some rust and contribute a feature. 

Any feedback welcome :) 

Todo:
- [x] Block outgoing connections 
- [ ] Mark peer as blocked 

Future work:
Convert IP ranges to CIDR notation and use a [prefix trie](https://docs.rs/prefix-trie/latest/prefix_trie/)
Save and load blocklists from disk and periodically pull blocklist url

### DIFF
--- a/crates/librqbit/src/blocklist.rs
+++ b/crates/librqbit/src/blocklist.rs
@@ -9,6 +9,7 @@ use tokio::io::AsyncRead;
 use tokio::{io::AsyncBufReadExt, io::BufReader};
 use tokio_util::io::StreamReader;
 use tracing::{debug, info, trace};
+use url::Url;
 
 pub struct Blocklist {
     // ipv4 and ipv6 do not overlap
@@ -28,7 +29,16 @@ impl Blocklist {
     }
 
     pub async fn load_from_url(url: &str) -> Result<Self> {
-        let response = reqwest::get(url)
+        let parsed_url = Url::parse(url).context("Failed to parse URL")?;
+
+        if parsed_url.scheme() == "file" {
+            let path = parsed_url
+                .to_file_path()
+                .map_err(|_| anyhow::anyhow!("Failed to convert file URL to path"))?;
+            return Self::load_from_file(path.to_str().unwrap()).await;
+        }
+
+        let response = reqwest::get(parsed_url)
             .await
             .context("Failed to send request for blocklist")?;
         if response.status() != 200 {
--- a/crates/librqbit/src/blocklist.rs
+++ b/crates/librqbit/src/blocklist.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use async_compression::tokio::bufread::GzipDecoder;
 use futures::TryStreamExt;
 use intervaltree::IntervalTree;
@@ -28,12 +28,11 @@ impl Blocklist {
     }
 
     pub async fn load_from_url(url: &str) -> Result<Self> {
-        let response = reqwest::get(url).await.map_err(|e| anyhow::anyhow!(e))?;
+        let response = reqwest::get(url)
+            .await
+            .context("Failed to send request for blocklist")?;
         if response.status() != 200 {
-            return Err(anyhow::anyhow!(
-                "Failed to fetch blocklist: HTTP {}",
-                response.status()
-            ));
+            anyhow::bail!("Failed to fetch blocklist: HTTP {}", response.status());
         }
 
         let content_length = response
@@ -41,9 +40,7 @@ impl Blocklist {
             .ok_or_else(|| anyhow::anyhow!("Failed to get content length"))?;
 
         if content_length < 2 {
-            return Err(anyhow::anyhow!(
-                "Content too short: not enough data to determine compression"
-            ));
+            anyhow::bail!("Content too short: not enough data to determine compression");
         }
 
         let reader = StreamReader::new(
@@ -72,9 +69,7 @@ impl Blocklist {
         if buffer.len() >= 2 {
             peek_bytes.copy_from_slice(&buffer[0..2]);
         } else {
-            return Err(anyhow::anyhow!(
-                "Content too short: not enough data to determine compression"
-            ));
+            anyhow::bail!("Content too short: not enough data to determine compression");
         }
 
         // Check for Gzip magic bytes (1F 8B)
--- a/crates/librqbit/src/blocklist.rs
+++ b/crates/librqbit/src/blocklist.rs
@@ -164,53 +164,12 @@ fn parse_ip_range(line: &str) -> Option<(IpAddr, IpAddr)> {
 
 #[cfg(test)]
 mod tests {
+    use std::io::Cursor;
+
     use super::*;
     use async_compression::tokio::write::GzipEncoder;
-    use mockito::{Server, ServerGuard};
-    use std::thread::{self, JoinHandle};
+    use futures::stream::once;
     use tokio::io::AsyncWriteExt;
-
-    struct TestServer {
-        server: ServerGuard,
-        mock: mockito::Mock,
-        url: String,
-        _thread: JoinHandle<()>,
-    }
-
-    impl TestServer {
-        fn new(content: &[u8], headers: &[(&str, &str)]) -> Self {
-            let (tx, rx) = std::sync::mpsc::channel();
-            let server_thread = thread::spawn(move || {
-                let mut server = Server::new();
-                let url = server.url();
-                let mock = server.mock("GET", "/").with_status(200);
-
-                tx.send((server, mock, url)).unwrap();
-                thread::park();
-            });
-
-            let (server, mut mock, url) = rx.recv().unwrap();
-
-            mock = mock.with_body(content);
-            for &(key, value) in headers {
-                mock = mock.with_header(key, value);
-            }
-            let mock = mock.create();
-
-            TestServer {
-                server,
-                mock,
-                url,
-                _thread: server_thread,
-            }
-        }
-    }
-
-    impl Drop for TestServer {
-        fn drop(&mut self) {
-            self._thread.thread().unpark();
-        }
-    }
 
     #[tokio::test]
     async fn test_blocklist_gzipped() -> Result<()> {
@@ -227,13 +186,13 @@ mod tests {
             encoder.shutdown().await.unwrap();
         }
 
-        let server = TestServer::new(&gzipped_blocklist, &[("Content-Encoding", "gzip")]);
-
-        let blocklist = Blocklist::load_from_url(&server.url).await?;
+        let stream = StreamReader::new(Box::pin(once(async {
+            Ok::<_, std::io::Error>(Cursor::new(gzipped_blocklist))
+        })));
+        let blocklist = Blocklist::create_from_stream(stream).await?;
         assert!(blocklist.is_blocked(&"192.168.1.1".parse().unwrap()));
         assert!(!blocklist.is_blocked(&"8.8.8.8".parse().unwrap()));
 
-        server.mock.assert();
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


crates/librqbit/src/session.rs
```diff
@@ -607,6 +612,14 @@ impl Session {
 
             let stream_connector = Arc::new(StreamConnector::from(proxy_config));
 
+            let blocklist: blocklist::Blocklist = if let Some(blocklist_url) = opts.blocklist_url {
+                blocklist::Blocklist::load_from_url(&blocklist_url)
+                    .await
+                    .unwrap()
```
ikatson, line 618: use anyhow's .context(...) here

crates/librqbit/src/session.rs
```diff
@@ -716,6 +730,11 @@ impl Session {
             .read_write_timeout
             .unwrap_or_else(|| Duration::from_secs(10));
 
+        let incoming_ip = addr.ip();
+        if self.blocklist.is_blocked(&incoming_ip) {
```
ikatson, line 734: only incoming ips are checked, how about outgoing

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+        ipv4_ranges: &Vec<std::ops::Range<u32>>,
+        ipv6_ranges: &Vec<std::ops::Range<u128>>,
+    ) -> Self {
+        Self {
+            ipv4_ranges: IntervalTree::from_iter(ipv4_ranges.iter().map(|r| (r.clone(), ()))),
+            ipv6_ranges: IntervalTree::from_iter(ipv6_ranges.iter().map(|r| (r.clone(), ()))),
+        }
+    }
+
+    pub async fn load_from_url(url: &str) -> Result<Self> {
```
ikatson, line 37: how about treating file:// urls specially?

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+use std::pin::Pin;
+use std::str::FromStr;
+use tokio::io::AsyncRead;
+use tokio::{io::AsyncBufReadExt, io::BufReader};
+use tokio_util::io::StreamReader;
+use tracing::{debug, info, trace};
+
+pub struct Blocklist {
+    // Separate trees for IPv4 and IPv6 since they have different numeric ranges
+    ipv4_ranges: IntervalTree<u32, ()>,
```
ikatson, line 15: I think you can store Ipv4Addr/Ipv6Addr everywhere here as they implement Ord, and that's the only necessary thing for both std::ops::Range and IntervalRange.

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+impl Blocklist {
+    pub fn empty() -> Self {
+        return Self::new(
+            &Vec::<std::ops::Range<u32>>::new(),
+            &Vec::<std::ops::Range<u128>>::new(),
+        );
+    }
+
+    pub fn new(
+        ipv4_ranges: &Vec<std::ops::Range<u32>>,
```
ikatson, line 28: this can be simplified like this:

```
pub fn new(
    ipv4_ranges: impl IntoIterator<Item=std::ops::Range<Ipv4Addr>>,
    ...
) -> Self {
    ipv4_ranges: IntervalTree::from_iter(ipv4_ranges.map(|r| (r, ())),
    ...
```

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            // Parse IP ranges in format: "RuleName:StartIp-EndIp"
+            if let Some((start_ip, end_ip)) = parse_ip_range(&line) {
+                match (start_ip, end_ip) {
+                    (IpAddr::V4(start), IpAddr::V4(end)) => {
+                        let start_num = u32::from_be_bytes(start.octets());
+                        let end_num = u32::from_be_bytes(end.octets());
+                        let range = if end_num == u32::MAX {
+                            start_num..end_num // Special case: Use inclusive range when max
```
ikatson, line 115: Can you just use inclusive ranges to simplify this?

```
start_num..=end_num
```

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+            }
+        }
+
+        info!(
+            ipv6_entry_count = ipv6_ranges.len(),
+            ipv4_entry_count = ipv4_ranges.len(),
+            "Finished loading blocklist"
+        );
+
+        let blocklist = Self::new(&ipv4_ranges, &ipv6_ranges);
```
ikatson, line 144: you can pass by value and consume, see also above comment with IntoIterator

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,339 @@
...
+                let num = u128::from_be_bytes(ipv6.octets());
+                self.ipv6_ranges.query_point(num).next().is_some()
+            }
+        }
+    }
+}
+
+fn parse_ip_range(line: &str) -> Option<(IpAddr, IpAddr)> {
+    // Skip comments and empty lines
+    if line.starts_with('#') || line.trim().is_empty() {
```
ikatson, line 164: how about

```
let line = line.trim();  // or maybe trim_end()
if line.starts_with('#') || line.is_empty()
```

so that it's more bulletproof against trailing whitespace

ikatson (changes requested):
Looks good, left a few comments

How widely used / popular are these blocklists? My only concern is increased number of dependencies and compilation time for a very rarely used feature

crates/librqbit/src/blocklist.rs
ikatson: Why?

xobust: IntervalTree [Nodes](https://docs.rs/intervaltree/latest/intervaltree/struct.Element.html) uses Range explicitly. 
So the inclusive range will need to be converted before it is constructed (?).

A better structure for the block list would be a [prefix trie](https://docs.rs/prefix-trie/latest/prefix_trie/index.html) but then a conversion to CIDR representation is needed. If the main concern is not fetching additional dependencies then I can make a naive implementation with existing data structures.

ikatson: Ah, makes sense. It's fine to leave the crate then.

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    pub fn new(ip_ranges: impl IntoIterator<Item = std::ops::Range<IpAddr>>) -> Self {
+        Self {
+            blocked_ranges: IntervalTree::from_iter(ip_ranges.into_iter().map(|r| (r, ()))),
+        }
+    }
+
+    pub async fn load_from_url(url: &str) -> Result<Self> {
+        let response = reqwest::get(url).await.map_err(|e| anyhow::anyhow!(e))?;
```
ikatson, line 31: you can probably just use .context() here and below.
Also anyhow::bail! is handier than "return Err(anyhow::anyhow!"

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+        let mut lines = reader.lines();
+        let mut ip_ranges: Vec<std::ops::Range<IpAddr>> = Vec::new();
+        while let Some(line) = lines.next_line().await? {
+            // Skip comments and empty lines
+            if line.starts_with('#') || line.trim().is_empty() {
+                continue;
+            }
+
+            if let Some((start_ip, end_ip)) = parse_ip_range(&line) {
+                let range = start_ip..(increment_ip(end_ip).unwrap());
```
ikatson, line 101: rather than panicng (which we should never do in librqbit), I think it's just safe to return "end_ip" as the range bound if it couldn't be incremented.
So that  you can just write ```let range = start_ip..increment_ip(end_ip)```

For both ipv4 and ipv6 those ips are either broadcast or reserved, so we shouldn't be connecting to them anyway

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+        info!(
+            ip_entry_count = ip_ranges.len(),
+            "Finished loading blocklist"
+        );
+
+        let blocklist = Self::new(ip_ranges);
+        Ok(blocklist)
+    }
+
+    pub fn is_blocked(&self, ip: &IpAddr) -> bool {
```
ikatson, line 115: nit: I'd pass IpAddr by value. Passing references for such primitive structs just adds a bit of syntax clutter without any benefit.

I.e. you need to write at least 3 more symbols (reference in caller(s), reference in signature, dereference below in "query_point") for no reason.

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.blocked_ranges.query_point(*ip).next().is_some()
+    }
+}
+
+/// Safely increments an `IpAddr`, returning `None` if it would overflow.
+fn increment_ip(ip: IpAddr) -> Option<IpAddr> {
+    match ip {
+        IpAddr::V4(ipv4) => {
+            let num = u32::from_be_bytes(ipv4.octets());
+            num.checked_add(1).map(|n| IpAddr::V4(Ipv4Addr::from(n)))
```
ikatson, line 125: given the above comment you might just use saturating_add and remove Option in result

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+fn parse_ip_range(line: &str) -> Option<(IpAddr, IpAddr)> {
+    // Skip comments and empty lines
+    let line = line.trim();
+    if line.starts_with('#') || line.is_empty() {
+        return None;
+    }
+
+    let is_ipv4 = line.matches('.').count() >= 6;
+    // Find the split point based on whether it's IPv4 or not
+    let split_point: usize = if is_ipv4 {
```
ikatson, line 143: why is this different for ipv4 and ipv6?

why not just

```
if let Some(rule_name, range) = split_once(':') {
    if let Some(start, end) = range.split_once('-') {
        ...
    }
}
```

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+    use tokio::io::AsyncWriteExt;
+
+    struct TestServer {
+        server: ServerGuard,
+        mock: mockito::Mock,
+        url: String,
+        _thread: JoinHandle<()>,
+    }
+
+    impl TestServer {
```
ikatson, line 180: How about removing HTTP servers and just test parsing streams?
Just to decrease complexity of this test a bit and remove mockito which wasn't necessary so far?

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,314 @@
...
+            Box::pin(BufReader::new(GzipDecoder::new(reader)))
+        } else {
+            trace!("Plain text file detected.");
+            Box::pin(reader)
+        };
+
+        let reader = BufReader::new(reader);
+        let mut lines = reader.lines();
+        let mut ip_ranges: Vec<std::ops::Range<IpAddr>> = Vec::new();
+        while let Some(line) = lines.next_line().await? {
```
ikatson, line 94: lines() allocates every line. Instead you could use read_line() from AsyncBufReadExt and reuse one allocation

crates/librqbit/src/blocklist.rs
xobust: While testing against some real world block list i found out that some rules contained colon ()':') in the rule name.
This would cause the rule to fail parsing. 
I think we could safely ignore those rules though since they are not well formatted.

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    pub async fn load_from_url(url: &str) -> Result<Self> {
+        let parsed_url = Url::parse(url).context("Failed to parse URL")?;
+
+        if parsed_url.scheme() == "file" {
+            let path = parsed_url
+                .to_file_path()
+                .map_err(|_| anyhow::anyhow!("Failed to convert file URL to path"))?;
+            return Self::load_from_file(path.to_str().unwrap()).await;
```
ikatson, line 38: rather than using .unwrap() here it would be better to pass &Path (or AsRef<Path>) into load_from_file

In any case let's strive not use .unwrap() in librqbit unless it's 100% clear it will never fail

crates/librqbit/src/blocklist.rs
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        let response = reqwest::get(parsed_url)
+            .await
+            .context("Failed to send request for blocklist")?;
+        if response.status() != 200 {
```
ikatson, line 44: nit: I think reqwests has some shortcut like .status().is_success()

ikatson (approved):
LGTM, thanks for the PR! Left a couple minor comments

PR #351, opened 2025-03-25
Add example systemd service 

mortn:
Systemd service added to show deployment example

### DIFF
--- a/rqbit.service
+++ b/rqbit.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=RQbit
+After=network.target
+
+[Service]
+Type=simple
+Environment=RQBIT_HTTP_API_LISTEN_ADDR=0.0.0.0:3030
+Environment=RQBIT_DIR=/home/downloads/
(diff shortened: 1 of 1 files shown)


ikatson:
Sorry, but I don't think it belongs in the core repo. It's highly specific to the distro, the user's preferences etc.

mortn:
A bit late to read this, but anyways:
[SystemD](https://en.wikipedia.org/wiki/Systemd) is the standard for service management on any Linux distro for services not running as container. It's the opposite of "highly specific to the distro". Anyways, it's your repo.
Thank you for this app. Great work.

ikatson:
@mortn what I meant is that the configuration is specific to the distro, not that systemd is an unusual thing. 

You can run the server with just "rqbit server start /folder". This can be translated to systemd or anything similar so easily you won't need an example. But what should "/folder" be? What other flags should be set? What ports to listen on, where to store config, etc. All those can be answered only by the user or the distro taking some defaults.

PR #358, opened 2025-04-12
:arrow_up: (librqbit) Update dep requirement for url to compatible w/ current version

djmango:
Less strict than =, can use as dep in other projects :)


michakfromparis:
I second this one

ikatson:
Merged, thanks for the PR! There was some broken version, I guess should be fixed in upstream considering how much time passed

ikatson:
Yeah I was planning to release 9-beta.0 soon with uTP support but if necessary I can just release a new 8.* with this fix

PR #367, opened 2025-04-26
watch .magnet files too

FrancescElies:
Currently only `.torrent` files are being considered when watching a folder.

This PR adds extends that functionality to `.magnet` files too.

### DIFF
--- a/crates/librqbit/src/watch.rs
+++ b/crates/librqbit/src/watch.rs
@@ -4,6 +4,7 @@ use std::{
     sync::{Arc, Weak},
 };
 
+use crate::Magnet;
 use anyhow::{bail, Context};
 use buffers::ByteBuf;
 use librqbit_core::torrent_metainfo::torrent_from_bytes;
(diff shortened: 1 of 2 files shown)


crates/librqbit/src/watch.rs
```diff
@@ -109,23 +121,37 @@ fn watch_thread(
...
+                        warn!(?path, "magnet file invalid: {e:#}");
+                        None
+                    }
+                },
+                Some(_) | None => {
+                    trace!(?path, "ignoring path");
+                    None
+                }
+            })
+            .map(|(add_torrent, path)| {
```
ikatson, line 148: Instead of ```let _results = map().collect()``` (which allocates needlessly and creates an unused variable) you can use ether a for loop or [for_each](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each)

crates/librqbit/src/watch.rs
```diff
@@ -109,23 +121,37 @@ fn watch_thread(
...
-                Ok(add) => add,
-                Err(e) => {
-                    warn!(?path, "error validating torrent: {e:#}");
-                    continue;
+
+        let _results = ev
+            .paths
+            .iter()
+            .filter_map(|path| match path.extension() {
+                Some(x) if x.to_str() == Some("torrent") => match read_and_validate_torrent(path) {
```
ikatson, line 129: it would probably be a bit shorter an more efficient if you do

```rust
match path.extension().and_then(|e| e.to_str())? {
   "torrent" => ...,
   "magnet" => ...,
}
```

crates/librqbit/src/watch.rs
FrancescElies: makes sense, done

crates/librqbit/src/watch.rs
FrancescElies: right, probably you don't want `?` as if by any reason a file without extension would land in the watch folder you would return and not process the rest of files right?

Changed it to `match path.extension().and_then(|e| e.to_str()) { ...}`

ikatson: As it's part of filter_map, the ? would only apply to the callback, not the outer function, so it should work

FrancescElies: Ok learned something new, thanks, i will change that

FrancescElies: Oh I somehow I forgot that it was inside a closure, that's why, it returns from there.

Done

PR #370, opened 2025-05-08
fix: improve path traversal check

illfygli:
The previous implementation would falsely detect a path traversal in a filename such as "foo... bar".
This patch changes the check function so that it inspects `PathBuf` components instead of substrings.

I don't know much Rust so I hope this makes sense. I have test-driven it myself and verified that I can now add a torrent with a filename containing `...` which previously did not work.


ikatson:
Thanks for the improvement!

PR #374, opened 2025-06-02
Include crate version in client fingerprint portion of `peer_id`

av-gal:
Include `librqbit`'s version number in its peer ids by default. Allow applications to specify a custom client fingerprint, and introduce a function to generate Azereus-style fingerprints from a two byte string and version tuple. Also bump librqbit-core to 5.0, since this PR changes its public API.


ikatson:
This is great, thanks! I wanted to do this for a long time but never got to it. 

One minor thing: can you please change getrandom create to random, cause we use rand already everywhere.

Issue #381, opened 2025-06-06
Library crate to extract torrent info from the magnet URL


oooo-ps:
I'm looking for a library crate that can store a temporary `.bittorrent` file in memory as an object, allowing me to extract the torrent content in my aggregator.

I might be able to use the rqbit server by adding a magnet link in a paused state (without downloading the content).

Thanks for any tips!

ikatson:
librqbit should be able to do all of this out of the box. Try it out and let us know if you have any issues

oooo-ps:
Thanks, already found this example in the repository:
https://github.com/ikatson/rqbit/blob/main/crates/librqbit/examples/ubuntu.rs

`AddTorrentOptions` + `list_only` or `paused` is exactly what I want!

UPD.  if someone has a similar task, take a look at this aggregation script for the `aquatic` tracker, which uses the `rqbit` library to convert info-hashes to a torrents files, memory objects, Manticore full-text index, etc:
https://github.com/YGGverse/aquatic-crawler

oooo-ps:
@ikatson I found only one issue: when trying to disable DHT for the qBittorrent session and provide a list of UDP torrent trackers instead, the `.torrent` list gets stuck on this line:
https://github.com/YGGverse/aquatic-crawler/blob/main/src/main.rs#L93

My testing of the UDP tracker is running in the IPv6/Yggdrasil network, but I'm not sure if it's related. There may be a bug related to IPv6. Could someone test it with an Internet IPv6 connection, as my provider does not offer this?

_The torrent file downloads successfully when I load it with DHT enabled, so the network interface is working properly._

ikatson:
There may be some IPv6 issue, it's kind of second-tier to be frank, I don't have IPv6 to test it, esp in IPv6-only mode.
You can try running with trace logs (e.g. with env ```RUST_LOG=librqbit=trace``` call ```tracing_subscriber::fmt::init()``` in your main) and see what it prints. Maybe reduce the list of trackers to just 1 or 2 for logs not to be very spammy

ikatson:
>  when trying to disable DHT for the qBittorrent session

I assume you meant librqbit session right?

oooo-ps:
> I assume you meant librqbit session right?

Yes in the file above, I start new `librqbit` session like this:
``` rust
let session = librqbit::Session::new_with_opts(
            std::path::PathBuf::new(),
            SessionOptions {
                disable_dht: !argument.enable_dht,
[...]

PR #382, opened 2025-06-07
Respect configurable listen interface for UDP tracker #381

oooo-ps:
According to https://github.com/ikatson/rqbit/issues/381#issuecomment-2952742738

I'm new in `rqbit` but I think this implementation is correct.

I'm also not sure how to properly implement a multi-interface listener (in parallel mode) as it is typically done in other BitTorrent clients. At the top level, the daemon should run both IPv4 and IPv6 interfaces simultaneously.

P.s. this solution has not been tested; it is primarily a draft implementation. It also requires an update to the crate version upon merging, as it includes changes to method arguments.

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -628,14 +628,10 @@ impl Session {
                 blocklist::Blocklist::empty()
             };
 
-            let udp_tracker_client = UdpTrackerClient::new(
-                listen_result
-                    .as_ref()
-                    .map_or((std::net::Ipv4Addr::LOCALHOST, 0).into(), |l| l.addr),
-                token.clone(),
-            )
-            .await
-            .context("error creating UDP tracker client")?;
+            let udp_tracker_client =
+                UdpTrackerClient::new(listen_result.as_ref().map(|l| l.addr), token.clone())
+                    .await
+                    .context("error creating UDP tracker client")?;
 
             let session = Arc::new(Self {
                 persistence,
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -628,10 +628,14 @@ impl Session {
                 blocklist::Blocklist::empty()
             };
 
-            let udp_tracker_client =
-                UdpTrackerClient::new(listen_result.as_ref().map(|l| l.addr), token.clone())
-                    .await
-                    .context("error creating UDP tracker client")?;
+            let udp_tracker_client = UdpTrackerClient::new(
+                listen_result
+                    .as_ref()
+                    .map_or((std::net::Ipv4Addr::UNSPECIFIED, 0).into(), |l| l.addr),
+                token.clone(),
+            )
+            .await
+            .context("error creating UDP tracker client")?;
 
             let session = Arc::new(Self {
                 persistence,
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -628,10 +628,14 @@ impl Session {
                 blocklist::Blocklist::empty()
             };
 
-            let udp_tracker_client =
-                UdpTrackerClient::new(opts.listen.unwrap_or_default().listen_addr, token.clone())
-                    .await
-                    .context("error creating UDP tracker client")?;
+            let udp_tracker_client = UdpTrackerClient::new(
+                listen_result
+                    .as_ref()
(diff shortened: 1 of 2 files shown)


oooo-ps:
I just found this namespace, which does not even contain IPv6:
``` rust
// crates/tracker_comms/src/tracker_comms_udp.rs
.. net::{Ipv4Addr, SocketAddr, SocketAddrV4}
```
It seems that this library works with IPv4 only and would require a significant refactor to handle IPv6 interactions (`parse_response` implementation for example)

@ikatson anyway, if you want to test IPv6, try installing [Yggdrasil](https://yggdrasil-network.github.io/) or [Mycelium](https://github.com/threefoldtech/mycelium) overlay networks. These are local networks, and the IPv6 addresses are 100% valid, which could be useful for testing connectivity over the IPv4 layer.

My local trackers there:
* `udp://[302:68d0:f0d5:b88d::fdb]:6969` - Yggdrasil
* `udp://[505:6847:c778:61a1::fdb]:6969` - Mycelium

oooo-ps:
Just wait, found this line above
``` rust
// crates/librqbit/src/session.rs
let listen_result = if let Some(listen_opts) = opts.listen.take() {
```
* so the `opts.listen` after this line will be always `None` > `default` > IPv4 / 0.0.0.0:0.

ikatson:
I'm not sure how this PR changes anything though?

I think you can try these:
1. listen on ```(Ipv6Addr::UNSPECIFIED, 0)```
2. use socket2 library like ```socket2::SockRef::from(&socket).set_only_v6(false)``` so that on Windows and MacOS it becomes dualstack (did not validate that this works). On Linux it's dualstack by default afaik

ikatson:
Can you try tracker-udp-dualstack branch?

oooo-ps:
I have this error:
``` bash
librqbit_tracker_comms::tracker_comms: error reading announce response: remote errored: Port can't be 0 url=udp://[302:68d0:f0d5:b88d::fdb]:6969
```
Branch connected with:
```Cargo.toml
[dependencies]
librqbit = {version = "9.0.0-beta.0", features = ["disable-upload"]}
[patch.crates-io]
librqbit = { git = "https://github.com/ikatson/rqbit.git", branch = "tracker-udp-dualstack", package = "librqbit" }
```

Also is this okay that following `SessionOptions` not available anymore?
1. `enable_upnp_port_forwarding`
2. `socks_proxy_url`

ikatson:
> Also is this okay that following SessionOptions not available anymore?

they were renamed, announcement here https://github.com/ikatson/rqbit/releases/tag/v9.0.0-beta.0

> remote errored: Port can't be 0 url=udp://[302:68d0:f0d5:b88d::fdb]:6969

Well, that's actually good - it's the response from your IPv6 tracker! Looking at the code, the port can be 0 in multiple scenarios:
- if session_opts.listen is None
- if list_only is True
- if the torrent is paused (probably not relevant here)

So try enabling TCP listener and not using list_only to see if you'll get further

ikatson:
Alternatively, I just pushed a commit where it defaults to 4240 instead of 0, so just try running again from the same branch

oooo-ps:
It seems I can't reset the Git cache for crate. I'll try again tomorrow. Thank you!

oooo-ps:
Changed my `Cargo.toml` to
``` Cargo.toml
librqbit = { path = "../../rqbit/crates/librqbit", package = "librqbit" }
```
Then changed `crates/tracker_comms/src/tracker_comms.rs`
``` rust
port: self.announce_port.unwrap_or(0)
```
to
``` rust
port: self.announce_port.unwrap_or(4240),
```
Now it works:
``` bash
Announce(AnnounceResponse { interval: 900, leechers: 0, seeders: 1, addrs: ...
```
Not sure, maybe you want global config instead of static `4240` value, or better - is port randomizing feature, or maybe some custom struct with the valid/shared `Default` implementation.

Also, yet another issue, btw, not related to one above
``` bash
DEBUG add_torrent: librqbit::dht_utils: error reading metainfo from [client::ipv6:xxx:xxx::]:12345: error reading handshake: peer disconnected while reading handshake
```
* DHT is disabled in my crawl configuration, maybe the Aquatic tracker issue

oooo-ps:
I think we can close this PR, by `tracker-udp-dualstack` branch implementation,
just make one patch with port change on example above, then we can marge on release.

PR #413, opened 2025-06-16
Fix shell completion output

Jeidnx:
I've noticed that when generating shell completions a log message is printed to the top of the output. This results in the completions not being usable.

This change runs the completions before anything else, resolving the issue.


ikatson:
Thanks for this! Never tried using it before, I guess it's a good time :)

PR #459, opened 2025-07-21
Update README for Homebrew install

rossmacarthur:
I added **rqbit** to Homebrew so you can now install it using
```
brew install rqbit
```


ikatson:
Wow, thanks!

Issue #473, opened 2025-08-10
How to set custom User agent?


oooo-ps:
Looking for librqbit API, if exists or just any keyword to find it.

<img width="177" height="79" alt="Image" src="https://github.com/user-attachments/assets/89ea5cb7-5f5d-4a65-b669-024072b9f44e" />

ikatson:
You can't. The code that generates this string is "client_name_and_version()"

oooo-ps:
Thanks for the tip, does it make sense to PR this option?

I also saw that some torrent clients [block alpha releases](https://github.com/transmission/transmission/issues/592#issuecomment-629732570). I’m not sure why; just a note.

P.S. I want this feature to indicate my crawler connection.

ikatson:
Feel free to send a PR to make it overridable

oooo-ps:
I see that the current implementation is using a zero-copy method to obtain the user agent value (`&str`). I'm not sure if it's a good idea to store the default value in `PeerConnectionOptions` because I don't know how many times it spawns.

Another idea to use an environment parameter is not optimal either, because it returns a `String` (which involves a copy).
``` rust
pub const fn client_name_and_version() -> &'static str {
    std::env::var("USER_AGENT").or(concat!("rqbit ", env!("CARGO_PKG_VERSION"))) // wants copy
}
```

Anyway, I would prefer to use an API rather than environment variables. Since `PeerConnectionOptions` requires a `Copy` implementation, perhaps I should store the settings at the parent level...

ikatson:
- Pass a String in SessionOptions, and store that String in Session
- Don't change client_name_and_version(). Maybe rename to default_client_name_and_version()
- Drill it down to PeerConnection somehow. One option is adding a method with a default implementation to PeerConnectionHandler trait (e.g. ```fn client_name_and_version(&self) -> &str```). Override that method in the relevant impls, and fetch that string from the session

PR #476, opened 2025-08-12
Custom user agent option

oooo-ps:
I see the #473 implementation like that
imho, the `crate::client_name_and_version` is no longer required by the current implementation.

However, I have no idea how to delegate its value to the `PeerConnection::manage_peer`

* based on the [PR#475](https://github.com/ikatson/rqbit/pull/475) branch

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -1,3 +1,5 @@
+mod user_agent;
+
 use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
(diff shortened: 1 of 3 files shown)


ikatson (changes requested):
1. please rebase on master, don't include the clippy changes here
2. you actually made it worse for the default case - you are allocating the default client_name_and_version into a String.
3. There's no point in the UserAgent(String) wrapper and a separate module for this, just use a String.

> However, I have no idea how to delegate its value to the PeerConnection::manage_peer

I described exactly how in https://github.com/ikatson/rqbit/issues/473

PR #492, opened 2025-10-05
add FTP storage backend

milahu:
early draft to add a FTP storage backend

use case:
bittorrent reverse proxy
running on a fast cache server with limited storage
in front of a slow backend server with large storage

`remotefs-ftp` does not work because `remotefs` is not thread-safe
see e9dcd1697cb8cbeaa4a4dd6c7c349d61a79955cf

`async_ftp` is hard to integrate
because all the storage functions are sync...
i am not good enough in Rust for such a large task
and github copilot fails on complex errors

note:
the FTP protocol is an implementation detail
it could be any network filesystem:
NFS, SMB, HTTP, WebDAV, SCP, Rsync, Rclone, ...

alternatives:

- mount the FTP server with [curlftpfs](https://sourceforge.net/projects/curlftpfs/) or [go-fuseftp](https://github.com/telepresenceio/go-fuseftp), but that requires the `fuse` kernel module
- hook syscalls with [fakechroot](https://github.com/dex4er/fakechroot) or [proot](https://github.com/proot-me/proot), but these tools have no FTP storage backend
- use a different bittorrent client like [libtorrent](https://github.com/arvidn/libtorrent) or [anacrolix/torrent](https://github.com/anacrolix/torrent), and add a FTP storage backend there
  - https://github.com/jpillora/cloud-torrent/issues/24


teodorkostov:
Pointless. Mount FTP to file system and run rqbit.

This PR will increase the codebase needlessly.

milahu:
> Mount FTP to file system and run rqbit

this should run on a shared server, so i have:
no root access, no fuse kernel module, no ptrace, ...

the first step to implement such a network storage backend would be
to convert all filesystem operations to async

teodorkostov:
If you are using it like that and you are facing such challenges it should occur to you that the technologies you are using are probably not right for the job. Storage for servers, VMs and orchestrators is a problem that has been already solved so many times. No point in adding code to a torrent client that would solve the same thing again.

Issue #495, opened 2025-10-14
The whitelist option [librqbit]


oooo-ps:
I want reverse feature for the `blocklist` one. Before PR, would ask: does blocklist.txt support some `!` (NOT) operator?
Maybe I can simply use the syntax in this case, without implementing extra API.

The goal - make outgoing connection possible to allowed peers only, by IP.

ikatson:
https://github.com/ikatson/rqbit/blob/main/crates/librqbit/src/blocklist.rs

There's obviously no negation support.

Feel free to create a PR for allowlist, here's some asks:
- to be consistent, name it "allowlist"
- all the code in blocklist.rs should be reused. Maybe rename the Blocklist struct and file to IpRanges or smth like that
- add CLI option with description --allowlist-url, similar to --blocklist-url

oooo-ps:
Just created PR, then found note no.1 - I want to rename, please wait a moment.
UPD. [done](https://github.com/ikatson/rqbit/pull/496/commits/816c0a03aa0fc500112547c5e994321b7baf10b7)

oooo-ps:
I'm just curious: should I restart the librqbit session to apply block or allow list updates? It seems that the registry is defining this anyway, so if the remote list has been changed, will these changes be applied 'on the fly'? Just lazy to open editor again.

The PR was merged, I'm closing this issue, thanks!

ikatson:
No, it won't get refreshed on the fly, you need to restart

oooo-ps:
If the list is not updating in session, maybe it's better to define the `blocklist` as `None`? Same as it was implemented for the `allowlist`, to not allocate extra memory.

Doubts only, maybe it could be useful to refresh this list in live mode, without session restart.. but remote source requires an additional validation on/and every new read, not sure that live syncing feature is really good idea.. So it could be better to init as `None` (in memory) if not requested by argument option, instead of creating empty list.

PR #496, opened 2025-10-14
implement whitelist option #495

oooo-ps:
#496

Please review, before merge as untested (only passing `cargo test`)

P.S. Why not to use the similar `Option` for the blacklist also, instead of empty struct. Is this updating dynamically somewhere? I think we may have ability to update list without app reload.


oooo-ps:
I think the list holder could be merged to some shared struct, to not copy similar features.
But this way requires API change.

PR #497, opened 2025-10-14
Allowlist option (based on the shared registry implementation)

oooo-ps:
Shared list implementation #495, https://github.com/ikatson/rqbit/pull/496

### DIFF
--- a/crates/librqbit/src/http_api/handlers/other.rs
+++ b/crates/librqbit/src/http_api/handlers/other.rs
@@ -70,7 +70,7 @@ pub async fn h_resolve_magnet(
     );
 
     if let Some(name) = info.name()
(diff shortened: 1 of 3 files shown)


oooo-ps:
also, applied clippy, last one: do you want it or I can remove by the way?

```
cargo clippy
warning: struct `GetPeersResponse` is never constructed
   --> crates/dht/src/bprotocol.rs:367:12
    |
367 | pub struct GetPeersResponse<BufT: ByteBufT> {
    |            ^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: `librqbit-dht` (lib) generated 1 warning
```
* https://github.com/ikatson/rqbit/pull/498

PR #498, opened 2025-10-14
require clippy + rustfmt for all targets

oooo-ps:
code contain warnings all the time, please require this test for PRs (all targets)


ikatson:
Thanks!

oooo-ps:
And thanks to you. After complete the check, we should catch this last warning (and fix it anyhow: remove or comment)
https://github.com/ikatson/rqbit/pull/497#issuecomment-3402442572

PR #499, opened 2025-10-14
Ip ranges init optimization

oooo-ps:
an addition to #497 includes minor `blocklist` init optimization described here: https://github.com/ikatson/rqbit/issues/495#issuecomment-3402773232
+ fixed `allowlist` environment name

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -140,7 +140,7 @@ pub struct Session {
     pub(crate) concurrent_initialize_semaphore: Arc<tokio::sync::Semaphore>,
     pub ratelimits: Limits,
 
-    pub blocklist: IpRanges,
+    pub blocklist: Option<IpRanges>,
     pub allowlist: Option<IpRanges>,
 
     // Monitoring / tracing / logging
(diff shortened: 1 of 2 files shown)


ikatson:
This won't actually optimize anything, and can make it worse even!

Wrapping in Option<...> will make it take more space, not less
All the callsites will get uglier too.
Yes, the init default() code won't be run but who cares - it's very cheap code for zero elements.

ikatson:
> fixed allowlist environment name

this one is good though, need to merge but only that, not the "optimisation". 
For allowlist having an Option makes sense because an empty allowlist vs None allowlist behave very differently.

oooo-ps:
Could you please [fix env var name](https://github.com/ikatson/rqbit/pull/499/commits/98260e49aff664f07fc8d3e28d214a6f2e34dc31) directly in main - I have no free memory to run IDE for right now.

ikatson:
> Could you please [fix env var name](https://github.com/ikatson/rqbit/pull/499/commits/98260e49aff664f07fc8d3e28d214a6f2e34dc31) directly in main

done

PR #505, opened 2025-10-31
Systemd socket activation

Stebalien:
*I'm submitting this early for feedback in case you aren't interested in maintaining/accepting this feature.*

This PR implements [systemd socket activation](https://0pointer.de/blog/projects/socket-activation.html) support (WIP). It makes it possible to auto-start rqbit (via systemd) when the user opens the webui, connects to the API, etc.

Implementation wise, this patch simply checks if the process has been passed a file descriptor by systemd and, if so, it uses that file descriptor as the API socket instead of opening a new one. This change doesn't add any new dependencies.

The main remaining TODO item is to implement a way to create a "dual stack" `TcpListener` from a file descriptor (requiring changes to https://github.com/ikatson/librqbit-dualstack-sockets).

The unit files (in the new `systemd` directory) are provided for documentation purposes and will likely need to be modified by the end-user (I'll add a README to the systemd directory later).


crates/rqbit/src/main.rs
```diff
@@ -908,9 +946,17 @@ async fn start_http_api(
         Some(log_config.rust_log_reload_tx),
         Some(log_config.line_broadcast),
     );
+    let systemd_listener = systemd_socket_activation().unwrap_or_else(|e| {
```
Stebalien: I'm happy to pass the socket into `start_http_api` instead, that'll just require a little more refactoring.

systemd/rqbit.conf
```diff
@@ -0,0 +1,36 @@
+#RQBIT_LOG_LEVEL_CONSOLE=
```
Stebalien: This is just a listing of all supported environment variables. It doesn't need to exist, it's just a nice way to let the user know what they can configure (I'll need to add documentation/defaults, this is just a first pass).

ikatson (commented):
Thanks for this, it certainly sounds useful so please go ahead and submit one this is ready

Stebalien:
The PR to librqbit-dualstack-sockets is up: https://github.com/ikatson/librqbit-dualstack-sockets/pull/3

Stebalien:
I've tested this locally against https://github.com/ikatson/librqbit-dualstack-sockets/pull/3 and it works.

Stebalien:
@ikatson when you get a chance, could you update `librqbit-dualstack-sockets` in `librqbit-utp`? That'll let me use that here and move this PR forward.

ikatson:
done and published utp 0.7.0

Stebalien:
This should now be ready:

- I put the documentation in the README itself.
- I removed the `.path` unit, but I'm happy to add some text to the readme describing how to auto-activate rqbit when `.torrent` files appear in their `~/Download` directory.

PR #506, opened 2025-11-01
Add buttons to pause and resume all torrents

fmckeogh:
Apologies, I don't really know what I'm doing with react.

It works, but it takes a second to update the torrents, I'm guessing there's some callback or something I need to be doing to trigger that correctly.

Could be done with two new HTTP endpoints, the current method requires `N+1` requests. 

I think a single button that toggles between pausing and resuming all would be nice.

### DIFF
--- a/crates/librqbit/webui/src/components/Header.tsx
+++ b/crates/librqbit/webui/src/components/Header.tsx
@@ -1,5 +1,7 @@
 import { FileInput } from "./buttons/FileInput";
 import { MagnetInput } from "./buttons/MagnetInput";
+import { PauseButton } from "./buttons/PauseButton";
+import { ResumeButton } from "./buttons/ResumeButton";
 
 // @ts-ignore
 import Logo from "../../assets/logo.svg?react";
@@ -23,6 +25,8 @@ export const Header = ({
         </h1>
       </div>
       <div className="flex flex-wrap gap-1 m-2">
+        <PauseButton className="flex-grow justify-center dark:text-white" />
+        <ResumeButton className="flex-grow justify-center dark:text-white" />
         <MagnetInput className="flex-grow justify-center dark:text-white" />
         <FileInput className="flex-grow justify-center dark:text-white" />
       </div>
--- a/crates/librqbit/webui/src/components/buttons/PauseButton.tsx
+++ b/crates/librqbit/webui/src/components/buttons/PauseButton.tsx
@@ -1,0 +1,40 @@
+import { useContext, useState } from "react";
+import {
+  ErrorDetails as ApiErrorDetails,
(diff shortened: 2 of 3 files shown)


fmckeogh:
I'm going to rewrite this with a proper endpoint, sending 200 requests is crazy and leaves it in a partially-paused state if interrupted.

Also if I worked on sort + search functionality for torrents would that be something you'd merge?

ikatson:
> Also if I worked on sort + search functionality for torrents would that be something you'd merge?

There's Compact UI branch, it has sorting. I'd start off there. You can slap on search onto it too probably.
I don't remember why I didn't merge it, I think just wanted to give it some rounds of testing / polishing.

As for mass start / mass pause - I think it's fine to do it with N requests, as they can be started / stopped independently, and will succeed at a different pace. Unless something goes wrong with this approach - I haven't tried, and looks like you did.

> sending 200 requests is crazy

What's so crazy about it? It's a bit suboptimal yes, but who cares? A bit of CPU wasted in JS runtime?

> leaves it in a partially-paused state if interrupted.

Interrupted by what? What is "partially paused"?

fmckeogh:
> There's Compact UI branch, it has sorting. I'd start off there. You can slap on search onto it too probably.

Awesome, thanks

> What's so crazy about it? It's a bit suboptimal yes, but who cares? A bit of CPU wasted in JS runtime?

It actually took 10+ seconds on server (i5 7200U) with ~500 torrents, client was my M2 MacBook Air. Not a super powerful server, but 10+ seconds was a lot. htop was showing rqbit pinned at 100% for that period.

> Interrupted by what? What is "partially paused"?

It was working fine, but I tried pausing once from my phone with spotty 4G and I couldn't get it to pause all torrents, only some. By "partially" I meant some torrents paused and others unpaused.

ikatson:
> Not a super powerful server, but 10+ seconds was a lot. htop was showing rqbit pinned at 100% for that period.

Oh I bet that's not the fact that there were a lot of requests. 500 requests is nothing for axum and even v8. But rather what is happening behind the scenes when you pause or start. Starting does a lot of stuff. Not sure if shoving all that into 1 request will make things better.

> and I couldn't get it to pause all torrents, only some. By "partially" I meant some torrents paused and others unpaused.

Well, in a case like this you want them to 
1) still get to the required state even if the client was disconnected
2) keep informing the client about the updates

You could do this on the server, and keep streaming the status to the client. But is it worth it just for 1) to work? I'd say no. And thus doing it on the client would be better. Maybe just throttle, don't try to start/pause more than N at a time.

ikatson:
This was done here https://github.com/ikatson/rqbit/pull/529

PR #510, opened 2025-11-03
Verify on-disk data when any pieces are missing, to avoid unnecessary re-downloading

fmckeogh:
I am seeing many of my `.bitv` files cleared (unsure as to the cause, no ZFS errors), which would be fine, except rqbit attempts to re-download the entire file, even though it already exists on disk. This is due to the `have_pieces` still being `Some`, so even though they are all `0`s but the full file exists on disk, we never execute the checksum validation codepath.

I believe the following code corrects this behavior.


fmckeogh:
I think this causes a UI regression however:(

Screenshot 2025-11-03 at 19 05 20

fmckeogh:
Haven't seen the UI regression again, I think this PR is good to go

PR #512, opened 2025-11-06
[WIP] Use workspace dependencies and try to unify duplicates

bbb651:
The main motivation is trying to resolve many heavy duplicated dependencies e.g. `tokio`, `sqlx-core`, `sqlx-postgres`, `serde_core`, `axum`, unfortunately it didn't seem to really work, what is really odd is that they all (except `axum`) seem to be using the same version and subset of each other's features which contradicts my mental modal of cargo.
I merged all `[dependencies]`, and then tries to merge

I experimented with different things:

- I added `tokio` without features to the workspace dependencies and then added the features explicitly to the dependencies in the crate, which seems better and probably should be done for all dependencies (and would avoid unnecessary features of dependencies when only using some of the crates in the workspace), see: https://github.com/rust-lang/cargo/issues/12162#issuecomment-2010322151
- I tried removing this [`tokio.features = ["all"]`](https://github.com/ikatson/rqbit/blob/8d99612023355076ca6439c37e88acb35bdd47bb/crates/upnp-serve/Cargo.toml#L14) which seems unnecessary by roughly guessing the features and seeing the errors and it seemed to work first try, but now that I think about it it's probably just using the shared tokio dependency and is probably missing some features
- Upgrading `console-subscriber` to `0.5.0` removed the duplicated `axum` dependency and massively reduced compile time from ~1m to ~37s
- I needed to remove `librqbit-sha1-wrapper.default-features = false` for it to pick the default backend and compile, not sure how it worked before

(P.S. I also sorted the dependencies but with vim `:sort` which doesn't match cargo notion of sorting with things like `cargo add` because of the order of some special characters like `-` and `_`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,38 +285,11 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.7.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
-dependencies = [
- "async-trait",
- "axum-core 0.4.5",
- "bytes",
- "futures-util",
- "http",
- "http-body",
- "http-body-util",
- "itoa",
- "matchit 0.7.3",
- "memchr",
- "mime",
- "percent-encoding",
- "pin-project-lite",
- "rustversion",
- "serde",
- "sync_wrapper",
- "tower 0.5.2",
- "tower-layer",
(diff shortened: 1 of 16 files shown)


ikatson:
Thanks for working on this!

> Upgrading console-subscriber to 0.5.0 removed the duplicated axum dependency and massively reduced compile time from ~1m to ~37s

console-subscriber is an optional depdendency afaiu though, so it shouldn't have mattered?

PR #531, opened 2026-01-13
Network Connectivity Improvements (Add Announce Port Override And ipv4-only Mode)

alexmichaelkeith:
I think this project is fantastic! To optimize my own experience, I implemented these changes in my fork, but I believe others could benefit from them as well (and quite frankly, I'd love to avoid maintaining a permanent fork :sweat_smile:).

This PR addresses connectivity issues encountered when running rqbit behind certain VPN providers (in my case ProtonVPN on macOS) and in environments lacking IPv6 support. 

### 1. Announce Port Override: _(Most Important to me)_
Many VPN services, including ProtonVPN, assign an external port for port forwarding that differs from the internal listening port. For example, rqbit might bind locally to port 4240, but the VPN provider forwards external traffic from port 55555 to 4240. Without the ability to manually specify the announce port, rqbit advertises its local port (4240) to trackers and the DHT. External peers attempt to connect to that port and fail because the VPN is only forwarding traffic on the assigned external port (55555).

### 2. IPv4-Only Mode _(Moderately important to me)_
Some network configurations, including the ProtonVPN GUI client (at least on macOS), often lack full IPv6 support or have inconsistent IPv6 connectivity. In these cases:
- The client may waste resources attempting to connect to IPv6 peers that are unreachable.
- The client might erroneously announce an invalid or unreachable IPv6 address to trackers.
- Logs become flooded with connection errors related to IPv6 reachability.

### 3. Port 0 Filtering _(Least Important to me)_
Added explicit filtering to reject peers advertising port 0, which is technically invalid and often indicative of "junk" data or bad NAT mappings, further improving connection stability.

### Changes

- Announce Port Override: Added an announce_port option to ListenerOptions and exposed it via the --announce-port CLI flag. This allows users to correctly advertise their external (VPN-assigned) port to the swarm while keeping the local bind port unchanged.
- IPv4-Only Mode: Added an ipv4_only mode (via --ipv4-only CLI flag). When enabled:
    - The listener strictly binds to IPv4 interfaces.
    - IPv6 peer addresses are filtered out during connection attempts.
    - The DHT explicitly patches the listen address to 0.0.0.0 instead of [::] to ensure consistent behavior.
- Port 0 Filtering: Added explicit filtering to reject peers advertising port 0, which is technically invalid and often indicative of "junk" data or bad NAT mappings, further improving connection stability.

### DIFF
--- a/crates/librqbit/examples/simulate_traffic.rs
+++ b/crates/librqbit/examples/simulate_traffic.rs
@@ -151,6 +151,8 @@ impl TestHarness {
                     listen_addr: (Ipv6Addr::UNSPECIFIED, 0).into(),
                     enable_upnp_port_forwarding: false,
                     utp_opts: None,
+                    announce_port: None,
+                    ipv4_only: false,
                 }),
                 ratelimits: LimitsConfig {
                     upload_bps: NonZero::new(64 * 1024),
@@ -235,6 +237,8 @@ impl TestHarness {
                     listen_addr: (Ipv6Addr::UNSPECIFIED, 0).into(),
                     enable_upnp_port_forwarding: false,
                     utp_opts: None,
+                    announce_port: None,
+                    ipv4_only: false,
                 }),
                 disable_local_service_discovery: false,
                 connect: Some(ConnectionOptions {


ikatson:
Hi, thanks a lot for this PR, looks great, and all the features are welcome! I'd merge right away just please fix cargo fmt

alexmichaelkeith:
Sweet! Should be formatted and up to date with main. If there is anything else needed to merge, just let me know. Thanks!

alexmichaelkeith:
Tests should pass now

PR #568, opened 2026-02-12
BEP 52 (2/16): Merkle tree module

meatsquirk:
## Summary

Implements PR 2 from `docs/BEP52-INCREMENTAL-PR-PLAN.md` (Merkle Tree Module).

- Add new `crates/librqbit_core/src/merkle.rs` with BEP 52 merkle primitives:
  - constants: `MERKLE_BLOCK_SIZE`
  - typed errors: `MerkleError`
  - core hashing: `zero_hash()`, `hash_block()`, `hash_pair()`
  - construction: `compute_merkle_root()` returning `MerkleResult { root, piece_hashes }`
  - verification: `verify_piece()`, `verify_block_with_proof()`
  - helpers: `padding_piece_hash()`, `root_from_piece_layer()`
- Export the module from `crates/librqbit_core/src/lib.rs` (feature-gated with sha1 backends).
- Include runtime correctness guards in verification paths:
  - reject oversized piece data in `verify_piece()`
  - reject out-of-range block indices in `verify_block_with_proof()`
- Implement small-tree stack path optimization (`padded leaf count <= 8`) while keeping Vec path for larger trees.
- Add broad unit coverage (33 merkle tests), including proof ordering, cross-validation, trimming behavior, and invalid input cases.

## BEP 52 Roadmap

This is PR 2 of 16 implementing BEP 52 (BitTorrent v2) support.
See https://gist.github.com/meatsquirk/533cae01ad9ba49834ffbb3bb3edc469 for the full stacked plan.

## Test plan

- [x] `cargo +1.90.0 check --all-targets`
- [x] `cargo +1.90.0 clippy --all-targets -- -D warnings -D dead_code -D unused_variables`
- [x] `cargo +1.90.0 test -p librqbit-core merkle`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Latest Update (2026-02-15)

Follow-up commit `217e3488` applies review-driven improvements to `merkle.rs`:

- Streaming merkle construction API: `compute_merkle_root_streaming()`.
- Single-pass piece verification (`verify_piece`) with no double chunk iteration.
- Tree reduction helper reuse to avoid duplicated pair-hashing loops.
- Small-tree stack path in `padding_piece_hash()`.
- Branchless fixed-size hash comparison (`ct_eq_id32`) in verification paths.
- Additional tests for streaming parity and constant-time comparison helper.

Additional validation run after this update:

- `cargo check`
- `cargo clippy --all-targets`
- `cargo test -p librqbit-core merkle::tests::`
- `cargo test --workspace`

All passed.

### DIFF
--- a/crates/librqbit_core/src/merkle.rs
+++ b/crates/librqbit_core/src/merkle.rs
@@ -17,6 +17,7 @@ use sha1w::{ISha256, Sha256};
 /// Size of each merkle tree leaf block (16 KiB), per BEP 52.
 pub const MERKLE_BLOCK_SIZE: usize = 16384;
 const SMALL_TREE_MAX_LEAVES: usize = 8;
+const MAX_TREE_LEVELS: usize = u32::BITS as usize;
 
 /// Errors from merkle tree operations.
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
@@ -67,6 +68,16 @@ pub fn hash_pair(left: &Id32, right: &Id32) -> Id32 {
     Id32::new(hasher.finish())
 }
 
+#[inline]
+fn ct_eq_id32(left: &Id32, right: &Id32) -> bool {
+    // Keep comparison branchless over all bytes.
+    let mut diff = 0u8;
+    for i in 0..32 {
+        diff |= left.0[i] ^ right.0[i];
+    }
+    diff == 0
+}
+
 fn validate_blocks_per_piece(blocks_per_piece: u32) -> Result<(), MerkleError> {
     if blocks_per_piece == 0 {
         return Err(MerkleError::ZeroBlocksPerPiece);
@@ -75,6 +86,24 @@ fn validate_blocks_per_piece(blocks_per_piece: u32) -> Result<(), MerkleError> {
         return Err(MerkleError::NonPowerOfTwoBlocksPerPiece(blocks_per_piece));
     }
     Ok(())
+}
+
+fn reduce_tree_in_place(current: &mut Vec<Id32>) {
+    debug_assert!(!current.is_empty());
+    debug_assert!(current.len().is_power_of_two());
+
+    let mut len = current.len();
+    while len > 1 {
+        for i in 0..(len / 2) {
+            // Read the current level from the upper range and write parent
+            // hashes into the lower range of the same buffer.
+            let left = current[i * 2];
+            let right = current[i * 2 + 1];
+            current[i] = hash_pair(&left, &right);
+        }
+        len /= 2;
+        current.truncate(len);
+    }
 }
 
 /// Build a balanced binary tree from power-of-2 leaves and return the root.
@@ -87,13 +116,7 @@ fn build_subtree_root_vec(leaves: &[Id32]) -> Id32 {
     }
 
     let mut current = leaves.to_vec();
-    while current.len() > 1 {
-        let mut next = Vec::with_capacity(current.len() / 2);
-        for pair in current.chunks_exact(2) {
-            next.push(hash_pair(&pair[0], &pair[1]));
-        }
-        current = next;
-    }
+    reduce_tree_in_place(&mut current);
     current[0]
 }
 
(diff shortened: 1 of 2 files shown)


ikatson:
Can we restructure the approach, so that there's no unused code?
For example, you could start with torrent metainfo / info parsing that would parse v2 fields.
Then progressively implement new things for v2 top-down (info parsing -> optionally fetching piece layers if we don't have them (magnet link case) -> validating data on disk -> downloading the missing pieces).

crates/librqbit_core/src/merkle.rs
```diff
@@ -0,0 +1,826 @@
...
+    }
+    build_subtree_root_vec(leaves)
+}
+
+/// Compute the full merkle tree from block hashes, extracting piece-layer hashes.
+///
+/// The tree is padded to a power-of-2 number of leaves with [`zero_hash()`].
+/// The piece layer is extracted at height `log2(blocks_per_piece)` from the leaves.
+/// Trailing piece hashes beyond the actual file content are trimmed.
+pub fn compute_merkle_root(
```
ikatson: Correct me if I'm wrong, but I see 2 use-cases for this:

1. creating a new torrent file (fairly low priority)
2. downloading a file

In both of these storing all the block hashes in memory first seems to be unnecessary memory overhead. Shouldn't we do both cases lazily?

All this maps into the main comment in the PR - I want to see this used first, otherwise it's just dead code that leaves more questions than answers.

crates/librqbit_core/src/merkle.rs
```diff
@@ -0,0 +1,826 @@
...
+    let root = build_subtree_root(&leaves);
+    Ok(root == *expected_hash)
+}
+
+/// Verify a single block using a merkle proof (bottom-up sibling hashes).
+///
+/// Proof ordering is bottom-up: `proof[0]` is the sibling at the leaf level,
+/// `proof[1]` at the next level up, etc. `block_index_in_piece` determines
+/// left/right placement at each level (bit 0 = leaf level, bit 1 = next, ...).
+pub fn verify_block_with_proof(
```
ikatson: mapping to the main PR comment, can we add all these functions ONLY when they get their first use

crates/librqbit_core/src/merkle.rs
meatsquirk: Correct use cases, we can switch to lazy/streaming approach, i made these changes and i'll update the pr.

meatsquirk:
I'll rework the plan to completely avoid dead code and try to vertically slice as much as possible incrementally adding functionality and minimizing risk to existing code.

meatsquirk:
Update for PR2 after review feedback (commit `217e3488`):

- Added `compute_merkle_root_streaming()` and kept `compute_merkle_root()` as a compatibility wrapper over `iter().copied()`.
- Reduced allocation churn in tree reduction by reusing a single buffer (`reduce_tree_in_place`) and removed duplicated pair-hashing loops.
- Made `verify_piece()` single-pass over piece bytes (`len().div_ceil()` for block count + streaming leaves), avoiding a second chunk iteration.
- Added no-allocation root reduction path for power-of-two leaves used by `verify_piece()`.
- Added branchless fixed-size hash comparison helper (`ct_eq_id32`) and switched verification equality checks to use it.
- Optimized `padding_piece_hash()` for small trees (`<= 8` leaves) to use stack-backed path.
- Added/updated tests:
  - `test_ct_eq_id32`
  - `test_compute_merkle_root_streaming_matches_slice_api`
  - `test_compute_merkle_root_streaming_without_hash_vec`

Validation run after these changes:

- `cargo check`
- `cargo clippy --all-targets`
- `cargo test -p librqbit-core merkle::tests::`
- `cargo test --workspace`

All passed.

ikatson:
As I said already, let's do other parts first, I want to see this used first before I spend time reviewing it again. It's all dead code, so let's do it only when it's actually needed (and I guess we won't need all of it, at least on the first use)

meatsquirk:
Closing per your feedback — I've reworked the plan into vertical slices with no dead code, following the top-down order you suggested (metainfo parsing first). The merkle code from this PR will land incrementally with its first consumers: the verification subset alongside piece-layer validation at add time, and the construction subset alongside v2 torrent creation. First slice (parse and validate v2/hybrid metainfo) coming shortly.

PR #577, opened 2026-03-31
Add configurable client identity for HTTP User-Agent and peer handshakes

libalpm64:
Many private trackers require an explicit User-Agent to be passed (to prevent crawling). UNIT3D-Announce (https://github.com/Roardom/UNIT3D-Announce) is commonly used on private trackers which will block any HTTP torrents from working when none is provided in the request header. Because of this change, I thought it might be fine to also add field overrides for the user-agent identifier as it was requested in #473. The implementation uses shallow copying which would be 1 string clone per torrent and 1 per peer connection (not really expensive). The added overhead is about 20 bytes.

### DIFF
--- a/crates/librqbit/src/lib.rs
+++ b/crates/librqbit/src/lib.rs
@@ -117,7 +117,7 @@ pub const fn version() -> &'static str {
 }
 
 pub const fn client_name_and_version() -> &'static str {
-    concat!("rqbit ", env!("CARGO_PKG_VERSION"))
+    concat!("rqbit/", env!("CARGO_PKG_VERSION"))
 }
 
 pub fn try_increase_nofile_limit() -> anyhow::Result<u64> {
--- a/crates/librqbit/src/lib.rs
+++ b/crates/librqbit/src/lib.rs
@@ -117,7 +117,7 @@ pub const fn version() -> &'static str {
 }
 
 pub const fn client_name_and_version() -> &'static str {
-    concat!("rqbit/", env!("CARGO_PKG_VERSION"))
+    concat!("rqbit ", env!("CARGO_PKG_VERSION"))
 }
 
 pub fn try_increase_nofile_limit() -> anyhow::Result<u64> {
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -1581,6 +1581,7 @@ impl Session {
             peer_rx,
             Some(self.merge_peer_opts(peer_opts)),
             self.connector.clone(),
+            self.client_name_and_version.clone(),
         )
         .await
         {
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -150,6 +150,7 @@ pub struct Session {
     _disable_upload: bool,
     pub ipv4_only: bool,
     pub peer_limit: Option<usize>,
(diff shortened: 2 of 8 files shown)


uplg:
Why not merged yet? This is crucial for any UNIT3D tracker in the wild.

libalpm64:
> Why not merged yet? This is crucial for any UNIT3D tracker in the wild.

Regarding your review I think we should keep the name/version as its the conventional UA format the space format is non standard practice and purely cosmetic.

crates/librqbit/src/lib.rs
```diff
@@ -117,7 +117,7 @@ pub const fn version() -> &'static str {
 }
 
 pub const fn client_name_and_version() -> &'static str {
-    concat!("rqbit ", env!("CARGO_PKG_VERSION"))
+    concat!("rqbit/", env!("CARGO_PKG_VERSION"))
```
uplg: This is wrong, should be space according to BEP 10 Extension Protocol but right for a classic HTTP user-agent

uplg:
> > Why not merged yet? This is crucial for any UNIT3D tracker in the wild.
> 
> Regarding your review I think we should keep the name/version as its the conventional UA format the space format is non standard practice and purely cosmetic.

The slash form is conventional for the HTTP User-Agent header specifically that's the RFC 9110 product/version shape and what tracker regexes expect on announces. But client_name_and_version() in this crate isn't (only) feeding that header. Its primary consumer is the BEP 10 extension handshake v field, and that protocol has its own de facto convention: a space.

  Look at what real clients send on the peer side:
  - µTorrent: µTorrent 3.5.5
  - qBittorrent: qBittorrent v4.6.0
  - Transmission: Transmission 3.00
  - Deluge: Deluge 2.1.1

  All space-separated. It's what the swarm "Client" column in every BT GUI parses today; flipping to rqbit/X.Y.Z would make rqbit the odd one out at the peer layer (cosmetic but visible to every peer we connect to). So the space form isn't arbitrary it's the existing peer-handshake convention, just as the slash form is the existing HTTP-UA convention.

  The cleanest split, I think, is:
  - Keep the const at rqbit X.Y.Z (space) — correct default for the peer handshake.
  - The new SessionOptions::client_name_and_version field lets a deployment override it for the HTTP UA case (where they want rqbit/X.Y.Z to please a strict tracker regex), without changing the wire identity peers see.

 That way both conventions stay correct for their respective protocol and downstream users pick the right format per use case rather than rqbit picking one and breaking the other.

crates/librqbit/src/lib.rs
libalpm64: Slash is not non-conformant but it doesn't matter but given pre-PR it was a space I think I will just revert it from a slash to a space for a seldom scenario.

uplg (approved):
LGTM !

crates/librqbit/src/session.rs
```diff
@@ -457,6 +458,10 @@ pub struct SessionOptions {
 
     /// Force IPv4 only.
     pub ipv4_only: bool,
+
+    /// Override the client name and version used in User-Agent headers and
+    /// peer extended handshakes. Defaults to "rqbit/X.Y.Z".
```
uplg: minor: deprecated comment

uplg (approved):

PR #580, opened 2026-04-15
fix: prevent from creating unselected files

exbit:


jasperweiss:
This would be very useful as I'm trying to use the library to download a single file from a torrent that contains hundreds/thousands of files.

```
2026-04-22T11:10:11.854566Z  INFO librqbit_dht::dht: DHT listening on 0.0.0.0:53394
2026-04-22T11:10:11.854575Z  INFO librqbit_dht::dht: starting up DHT with peer id ....
Error: add torrent for download

Caused by:
    0: error opening "..." in read/write mode
    1: Too many open files (os error 24)
```

ikatson:
This looks very reasonable on a quick look, thanks for this. I'll merge once I have a chance to test it out and study the details of math

exbit:
@ikatson Could you please review my request?

ikatson:
It looks a bit complex, how is this different from just lazily opening files on each .write_all() (and remove everything else in the change)?

PR #485, opened 2025-09-26
Allow specifying DHT bootstrap nodes from commandline or environment

crass:
Use the argument --dht-bootstrap-addrs or environment variable RQBIT_DHT_BOOTSTRAP to pass a comma separated list of ip:port or host:port values. This has been useful for me when the hostnames of the default bootstrap nodes to not resolve.

PR #583, opened 2026-05-06
feat: add DHT nodes from BEP-5 .torrent files to the routing table

chantra:
## Summary

Addresses the review on #583. BEP-5 `nodes` embedded in a `.torrent` are the closest DHT contacts, not BitTorrent peers — they must be queried over DHT to discover peers. The previous approach injected them into `initial_peers`, treating them as BT peers. This PR:

- Adds a public `DhtState::add_bootstrap_nodes(Vec<(String, u16)>)` API to inject nodes into the routing table at runtime (best-effort, background), resolving hostnames as well as IPv4/IPv6 literals (per the spec, `<host>` may be a DNS name like `your.router.node`).
- Refactors the existing `bootstrap_hostname` logic onto `DhtState` (`find_nodes_in_routing_table` + `resolve_and_find_nodes`) so bootstrap and the new API share one code path.
- Wires `.torrent` `nodes` in `librqbit`'s `add_torrent` to the DHT via that API instead of `initial_peers` (ignored with a debug log if DHT is disabled).

The initial BEP-5 commit was amended to remove the now-obsolete initial-peers wiring, keeping history clean.

## Test plan

- `cargo test -p librqbit-dht --lib` — new tests:
  - `bootstrap_populates_routing_table` (locks in existing bootstrap behavior)
  - `add_bootstrap_nodes_populates_table` (A learns a node only via the new API)
  - `add_bootstrap_nodes_empty_is_noop`
- `cargo test -p librqbit test_create_torrent_with_bootstrap_nodes`
- `cargo test -p librqbit-core dht_node::tests`
- `cargo clippy -p librqbit-dht --lib --tests` and `cargo clippy -p librqbit --all-targets` clean
- `cargo fmt --all -- --check` clean

Assisted by AI

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -1134,6 +1134,27 @@ impl Session {
                         }
                     };
 
+                    // BEP-0005: `nodes` are DHT contacts, not BT peers. Hand them
+                    // to the DHT so it can query them to discover peers. Host may
+                    // be a hostname or IP literal; the DHT resolves them.
+                    if !torrent.meta.nodes.is_empty() {
+                        match self.dht.as_ref() {
+                            Some(dht) => {
+                                let nodes: Vec<(String, u16)> = torrent
+                                    .meta
+                                    .nodes
+                                    .iter()
+                                    .map(|node| (node.host.clone(), node.port))
+                                    .collect();
+                                info!(?nodes, "BEP-0005: adding torrent nodes to DHT");
+                                dht.add_bootstrap_nodes(nodes);
+                            }
+                            None => {
+                                debug!("BEP-0005: torrent has nodes but DHT is disabled, ignoring");
+                            }
+                        }
+                    }
+
                     let mut trackers = torrent
                         .meta
                         .iter_announce()
--- a/crates/dht/src/dht.rs
+++ b/crates/dht/src/dht.rs
@@ -1396,3 +1396,80 @@ impl FromSocketAddr for SocketAddrV6 {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        net::{Ipv4Addr, SocketAddr},
+        time::Duration,
+    };
+
+    use librqbit_core::hash_id::Id20;
+
+    use crate::{Dht, DhtConfig, DhtState, routing_table::RoutingTable};
+
+    /// Spawn an in-process DHT node bound to an ephemeral v4 loopback port.
+    async fn spawn_node(
+        peer_id: Id20,
+        routing_table: Option<RoutingTable>,
+        bootstrap_addrs: Vec<String>,
+    ) -> Dht {
+        DhtState::with_config(DhtConfig {
+            peer_id: Some(peer_id),
+            bootstrap_addrs: Some(bootstrap_addrs),
+            routing_table,
+            listen_addr: Some((Ipv4Addr::LOCALHOST, 0).into()),
+            ..Default::default()
+        })
+        .await
+        .unwrap()
+    }
+
+    /// Poll the routing tables until a node with the given id shows up, or time out.
+    async fn wait_for_node_in_table(dht: &Dht, id: Id20, timeout: Duration) -> bool {
+        tokio::time::timeout(timeout, async {
+            loop {
+                let found = dht
+                    .with_routing_tables(|v4, v6| v4.iter().chain(v6.iter()).any(|n| n.id() == id));
+                if found {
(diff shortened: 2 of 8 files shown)


chantra:
@ikatson I am not super psyche about the current DhtNode serde, but did not find a good way to have it working using the existing bencode crate. Mostly due to [serialize_tuple](https://github.com/ikatson/rqbit/blob/f9b4aee85aff0fe52e206cfa3d3d5cc7e7d24947/crates/bencode/src/serialize.rs#L300-L302) . Let me know if there is ways you would prefer this written.

crates/librqbit/src/session.rs
```diff
@@ -1052,6 +1052,20 @@ impl Session {
                         }
                     };
 
+                    // BEP-0005: inject nodes from .torrent as initial peers.
+                    if !torrent.meta.nodes.is_empty() {
```
ikatson, line 1140: If I'm reading the spec correctly, these nodes aren't peers necessarily, but the closest DHT nodes.
Means they need to be queried through DHT to find peers, but not (just) added to initial parts

crates/librqbit/src/session.rs
```diff
@@ -1052,6 +1052,20 @@ impl Session {
...
+                        let node_peers: Vec<SocketAddr> = torrent.meta.nodes.iter()
+                            .filter_map(|node| {
+                                format!("{}:{}", node.host, node.port).parse::<SocketAddr>().ok()
```
ikatson: You can avoid the allocation by parsing IpAddr and combining with port afterwards

ikatson (changes requested):

ikatson:
As to custom deserialize/serialize that's totally fine and how it's usually done anyway

crates/librqbit/src/session.rs
chantra: yes, those are nodes that can be queried over DHT. Are you asking because then they should be added to a separate variable?

chantra: I suppose in some ways, it is similar to #485 . Would you prefer that the initial torrent nodes are added to the `Session.dht`?  I have not found a way to insert a node in the DhtState struct. Anything I am missing?

chantra:
Re-requesting review in case this is needed to get it back on your radar. 
I just realized tests are failing, will address in parallel.

crates/librqbit/src/session.rs
chantra: @ikatson I have been spending a bit more time on this and I believe my test worked initially because it probably leaked to the `DHT_BOOTSTRAP`. 
As much as I can see, in its current state, there is no way tell the Session to add those nodes to the dht table, right?

ikatson: Yes, and I think you need to add support for this to make this PR useful.

chantra: let me put this into draft mode for now. Thanks for the feedback.

chantra:
hi, just an update, still want to work on this, but need to prioritize other stuff right now.

crates/librqbit/src/session.rs
chantra: @ikatson ok, finally got to the bottom of this PR.

It is composed of 5 commits.
- [feat: add BEP-0005 DHT bootstrap nodes support in .torrent files](https://github.com/ikatson/rqbit/pull/583/changes/91c978ca07151b3490f3d17adc1c5816ed51b812): this is the existing code from this very PR adding support for .torrent.
- [test(dht): add loopback harness and bootstrap routing-table test](https://github.com/ikatson/rqbit/pull/583/changes/6448c3ad92e60a81117906e498520db873774154): Adds a test to make sure dht bootstrapping work after a refactor in next diff
- [refactor(dht): hoist resolve + find_node onto DhtState](https://github.com/ikatson/rqbit/pull/583/changes/ff6308eb4b3c76beca4bf07c1ceadabc1796477c): the refactor
- [feat(dht): add DhtState::add_bootstrap_nodes](https://github.com/ikatson/rqbit/pull/583/changes/77fdfab83b294ca94b7ec87423e4a1d06eda8bd9): builds the public `DhtState::add_bootstrap_nodes()` using the previous refactor + tests
-  [feat(librqbit): route .torrent BEP-5 nodes into the DHT](https://github.com/ikatson/rqbit/pull/583/changes/640f425d44df69d5a8721f68ba54bb80e0848896): inject the nodes discovered through the .torrent file into the DHT.

PR #591, opened 2026-05-17
librqbit(dht): provide an option to set DhtConfig::listen_addr

chantra:
Summary: Currently, if not using `PersistentDhtConfig`, it is not possible to set the listen_addr of DhtConfig.

I feel this should rather be captured under a "dht_config: Option<DhtConfig>" field instead, but it would conflict with the existing `pub dht_config: Option<PersistentDhtConfig>` so this version does not attempt to change any of this.

Test Plan: `cargo test`

### DIFF
--- a/crates/dht/src/persistence.rs
+++ b/crates/dht/src/persistence.rs
@@ -9,7 +9,7 @@ use librqbit_dualstack_sockets::BindDevice;
 use serde::{Deserialize, Serialize};
 use std::fs::OpenOptions;
 use std::io::{BufReader, BufWriter};
-use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 use tokio_util::sync::CancellationToken;
@@ -20,12 +20,29 @@ use crate::peer_store::PeerStore;
 use crate::routing_table::RoutingTable;
 use crate::{Dht, DhtConfig, DhtState};
 
+/// Configuration for DHT persistence (periodic dump of routing table and peer store).
+/// When provided, the DHT state will be serialized to disk periodically.
 #[derive(Default)]
-pub struct PersistentDhtConfig {
+pub struct DhtPersistenceConfig {
+    /// How often to dump state. Defaults to 60s.
     pub dump_interval: Option<Duration>,
+    /// Path to the JSON file. Uses OS-specific default if None.
     pub config_filename: Option<PathBuf>,
-    pub port: Option<u16>,
-    pub ipv4_only: bool,
+}
+
+/// Compute the DHT listen address from the explicit/stored port preferences and
+/// the session-level `ipv4_only` flag.
+///
+/// The bind IP is derived solely from `ipv4_only` (`0.0.0.0` if true, `[::]`
+/// otherwise); any persisted IP is intentionally ignored. The port is chosen
+/// in priority order: explicit -> stored -> 0 (random).
+pub fn dht_listen_addr(port: Option<u16>, stored_port: Option<u16>, ipv4_only: bool) -> SocketAddr {
+    let ip: IpAddr = if ipv4_only {
+        Ipv4Addr::UNSPECIFIED.into()
+    } else {
+        Ipv6Addr::UNSPECIFIED.into()
+    };
+    SocketAddr::new(ip, port.or(stored_port).unwrap_or(0))
 }
 
 #[derive(Serialize, Deserialize)]
@@ -85,14 +102,16 @@ impl PersistentDht {
 
     #[inline(never)]
     pub fn create<'a>(
-        config: Option<PersistentDhtConfig>,
+        persistence_config: DhtPersistenceConfig,
+        port: Option<u16>,
+        ipv4_only: bool,
+        bootstrap_addrs: Option<Vec<String>>,
         cancellation_token: Option<CancellationToken>,
         bind_device: Option<&'a BindDevice>,
     ) -> BoxFuture<'a, anyhow::Result<Dht>> {
         async move {
-            let mut config = config.unwrap_or_default();
-            let config_filename = match config.config_filename.take() {
-                Some(config_filename) => config_filename,
+            let config_filename = match persistence_config.config_filename {
+                Some(f) => f,
                 None => Self::default_persistence_filename()?,
             };
 
@@ -112,20 +131,8 @@ impl PersistentDht {
                     match serde_json::from_reader::<_, DhtSerialize<RoutingTable, PeerStore>>(
                         reader,
                     ) {
-                        Ok(mut r) => {
+                        Ok(r) => {
                             info!(filename=?config_filename, "loaded DHT routing table from");
-                            if config.ipv4_only {
-                                // Force IPv4 if requested
-                                let port = r.addr.port();
-                                r.addr = SocketAddr::from(([0, 0, 0, 0], port));
-                            } else if r.addr.ip() == Ipv4Addr::UNSPECIFIED {
-                                warn!(
-                                    "patching DHT listen IP address for rqbit 9 upgrade: 0.0.0.0 -> [::]"
-                                );
-                                let port = r.addr.port();
-                                r.addr = SocketAddr::from((Ipv6Addr::UNSPECIFIED, port));
-                            }
-
                             Some(r)
                         }
                         Err(e) => {
@@ -146,23 +153,18 @@ impl PersistentDht {
                     }
                 },
             };
-            let (mut listen_addr, routing_table, peer_store) = de
-                .map(|de| (Some(de.addr), Some(de.table), de.peer_store))
+            let (stored_port, routing_table, peer_store) = de
+                .map(|de| (Some(de.addr.port()), Some(de.table), de.peer_store))
                 .unwrap_or((None, None, None));
 
-            if let Some(port) = config.port {
-                if let Some(ref mut addr) = listen_addr {
-                    addr.set_port(port);
-                } else {
-                    listen_addr = Some(SocketAddr::from((Ipv6Addr::UNSPECIFIED, port)));
-                }
-            }
+            let listen_addr = dht_listen_addr(port, stored_port, ipv4_only);
             let peer_id = routing_table.as_ref().map(|r| r.id());
 
             let dht_config = DhtConfig {
                 peer_id,
+                bootstrap_addrs,
                 routing_table,
-                listen_addr,
+                listen_addr: Some(listen_addr),
                 peer_store,
                 cancellation_token,
                 bind_device,
@@ -175,7 +177,7 @@ impl PersistentDht {
                 dht.cancellation_token().clone(),
                 {
                     let dht = dht.clone();
-                    let dump_interval = config
+                    let dump_interval = persistence_config
                         .dump_interval
                         .unwrap_or_else(|| Duration::from_secs(60));
                     async move {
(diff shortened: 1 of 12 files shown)


chantra:
@ikatson your thoughts are welcome in how to make supporting "DhtConfig" with only one field. I can't think of a way to do it without breaking backward compatibility.

ikatson:
It's fine to break, 9 isn't released anyway

crates/dht/src/persistence.rs
```diff
@@ -114,11 +117,7 @@ impl PersistentDht {
                     ) {
                         Ok(mut r) => {
                             info!(filename=?config_filename, "loaded DHT routing table from");
-                            if config.ipv4_only {
-                                // Force IPv4 if requested
-                                let port = r.addr.port();
-                                r.addr = SocketAddr::from(([0, 0, 0, 0], port));
-                            } else if r.addr.ip() == Ipv4Addr::UNSPECIFIED {
```
chantra, line 121: the behaviour here changes. My read is that before it would load up the SocketAddr from the persistence file.
If ipv4_only was set, it would force to `Ipv4Addr::UNSPECIFIED`. If `ipv4_only` was not set and `Ipv4Addr::UNSPECIFIED` was read from the file, it would upgrade to `Ipv6Addr::UNSPECIFIED` down below.

Now, there is no `ipv4_only` anymore, but if the `r.addr()` was not `Ipv4Addr::UNSPECIFIED` it would not get upgraded, otherwise, it would.

crates/dht/src/persistence.rs
```diff
@@ -146,23 +145,18 @@ impl PersistentDht {
...
                 .map(|de| (Some(de.addr), Some(de.table), de.peer_store))
                 .unwrap_or((None, None, None));
 
-            if let Some(port) = config.port {
-                if let Some(ref mut addr) = listen_addr {
-                    addr.set_port(port);
-                } else {
-                    listen_addr = Some(SocketAddr::from((Ipv6Addr::UNSPECIFIED, port)));
-                }
-            }
```
chantra, line 159: Then we arrive here.
If `config.port` was not set, we would reuse the `listen_addr` we read from the persistence file, not doing the v6 upgrade.
If `config.port` is set, we would update the port of the address we read from the file, or would bind to `[::]:port`

Now, we try to use the listen addr provided by the user, if None, we use what we read from the file.

I feel the outcome is quite similar, but provides more flexibility to the caller? Happy to align more with the original flow if this makes more sense.

crates/librqbit/src/session.rs
```diff
@@ -464,6 +477,36 @@ pub struct SessionOptions {
     pub client_name_and_version: Option<String>,
 }
 
+impl Default for SessionOptions {
```
chantra, line 484: The `Default` impl is now needed because `disable_dht` and `disable_dht_persistence` are now equivalent to `dht: Some(DhtSessionConfig::default()),`

crates/rqbit/src/main.rs
```diff
@@ -603,14 +604,28 @@ async fn async_main(mut opts: Opts, cancel: CancellationToken) -> anyhow::Result
...
+        } else {
+            Some(DhtPersistenceConfig::default())
+        };
+        Some(DhtSessionConfig {
+            bootstrap_addrs: dht_bootstrap_addrs,
+            listen_addr: None,
+            persistence,
+        })
+    };
+
```
chantra, line 624: Setting `dht` is now a bit more verbose than used to with independent fields, but maybe more explicit.

chantra (commented):
ok, so I streamlined DhtConfig with the help of LLM. There is quite some changes in function signature which the commit description explains how to port forward.

The change in general LGTM, but I am saying that without having to care about backward compatibility, so happy to hear what you think.

Comment inline with some chunk of code worth highlighting.

One of the core change is that now a `SocketAddr` needs to be passed, not just a port, this comes with some slight behaviour change.

ikatson:
I really like the refactor of multiple session options into one "dht" option!

There's some weirdness with addresses/ports/ipv4_only flag though. And if you are touching this, would be good to make this right at the same time.

It seems there should be some logic, possibly extracted into a function that handles all combinations of:
- explicit listen address requested
- ipv4_only set or not (this option on the session must be respected)
- stored address in the routing table different from the requested one
- suspected rqbit 8 upgrade

I never asked, but now thinking about how messy this logic is / would be - why would you need to set listen address in the first place? At least it was all simpler, although still messy, when the address was not configurable at all. 

For context, the port+ipv4 was added in a [PR](https://github.com/ikatson/rqbit/pull/531]

ikatson (changes requested):
Left a comment for further discussion

chantra:
> I never asked, but now thinking about how messy this logic is / would be - why would you need to set listen address in the first place? At least it was all simpler, although still messy, when the address was not configurable at all.

Seems in the same vein as #531 . I have clients that run under docker and scheduled by nomad. The client is natted by docker and the port dynamically assigned at startup time. This allows me to start the DHT using a port that I know is opened so the client's DHT is accessible from outside.
In residential network, this is typically taken care of by upnp, but in this case this is not available.

Re ipv4-only. I would have to double-check, but it is not clear it was ever wired to DHT other than when explicitely building the session with a specific DhtPersistentConfig. I think this is the only invocation of [DhtPersistentConfig](https://github.com/ikatson/rqbit/blob/61e189a9b37ab3ba678efa66586c2c1aabeed584/desktop/src-tauri/src/main.rs#L111-L114). And the `SessionOptions`'s ipv4_only is never wired in `DhtPersistentConfig` in https://github.com/ikatson/rqbit/blob/61e189a9b37ab3ba678efa66586c2c1aabeed584/crates/librqbit/src/session.rs#L572

> ipv4_only set or not (this option on the session must be respected)

would providing an explicit `SocketAddr` solve this? Either one provides an IPv4/0.0.0.0 or a v6/[::]?

> suspected rqbit 8 upgrade
In such case, the upgrade is to go from 0.0.0.0 -> [::], right?

> stored address in the routing table different from the requested one

What is typically stored in that file, from looking at the code, it seems we don't care about the actual listen IP and default to using wildcard, upgrade it to [::] unless `v4_only` is set. Same here, would encoding the intent in a `SocketAddr` enough? Having it default to `[::]:0`?

ikatson:
> This allows me to start the DHT using a port that I know is opened so the client's DHT is accessible from outside.

But there was "port" exposed for that (not full listen addr) in the config. If you keep the change to just port + ipv4_only, then it seems it'll both fix your use-case, make the options saner, and keep existing (maybe not working as you say) ipv4_only feature which can be fixed as part of this PR.

So in short:
- compute IpAddr from ipv4_only flag, ignore the stored one completely
- use port in this order explicit port -> stored port -> new random port

chantra:
> But there was "port" exposed for that (not full listen addr) in the config.

yes, port was available, but only for persistent dht config, not plain dht config.

The benefit of listen_addr over just port is that you can define ip:port instead of *:port, which can be used to restrict how widely the port can be accessed. I don't have a strong need for specifying an IP in my case, so no strong opinions on that.

ikatson:
Exactly, if no-one needs this let's keep it simple. 
Bind device is there for limiting if needed. 
Is there's an actual usecase ever where a single interface has multiple ips we can reconsider

ikatson:
> yes, port was available, but only for persistent dht config, not plain dht config.

Just to clarify, my suggestion is to fix it in this PR as you are changing the config options entirely

chantra:
Sure, that's what I understood.

crates/dht/src/persistence.rs
```diff
@@ -112,20 +131,8 @@ impl PersistentDht {
                     match serde_json::from_reader::<_, DhtSerialize<RoutingTable, PeerStore>>(
                         reader,
                     ) {
-                        Ok(mut r) => {
+                        Ok(r) => {
                             info!(filename=?config_filename, "loaded DHT routing table from");
```
chantra, line 135: is there any cases where this may not be ipv4/ipv6::unspecified? If not, then it seems it is much easier to decide on [::] or 0.0.0.0 based on the ipv4_only value.
The `warn!` is removed, I can add it back if needed, just wondering if this is worth the if/else gymnastic.

crates/dht/src/persistence.rs
```diff
@@ -146,23 +153,18 @@ impl PersistentDht {
...
                 .unwrap_or((None, None, None));
 
-            if let Some(port) = config.port {
-                if let Some(ref mut addr) = listen_addr {
-                    addr.set_port(port);
-                } else {
-                    listen_addr = Some(SocketAddr::from((Ipv6Addr::UNSPECIFIED, port)));
-                }
-            }
+            let listen_addr = dht_listen_addr(port, stored_port, ipv4_only);
```
chantra, line 160: this takes care of ientifying if [::] or 0.0.0.0 should be used, as well as selecting the port to use based on the "explicit port -> stored port -> new random port" logic.

crates/dht/src/persistence.rs
```diff
@@ -206,3 +208,64 @@ impl PersistentDht {
...
+        assert_eq!(addr.port(), 6881);
+    }
+
+    #[test]
+    fn explicit_overrides_stored_v4() {
+        let addr = dht_listen_addr(Some(6881), Some(12345), true);
+        assert_eq!(addr.ip(), IpAddr::from(Ipv4Addr::UNSPECIFIED));
+        assert_eq!(addr.port(), 6881);
+    }
+}
```
chantra, line 271: probably a good candidate for table-driven tests. I have used rstest in the past for that. Lemme now if you are fine with the added dependency.

crates/librqbit/src/session.rs
```diff
@@ -556,30 +603,34 @@ impl Session {
...
-            let dht = if opts.disable_dht {
-                None
-            } else {
-                let dht = if opts.disable_dht_persistence {
+            let dht = if let Some(dht_config) = opts.dht.take() {
+                let dht = if let Some(persistence_config) = dht_config.persistence {
+                    PersistentDht::create(
+                        persistence_config,
+                        dht_config.port,
+                        opts.ipv4_only,
```
chantra, line 611: this was missing before, it is now wired.

crates/librqbit/src/session.rs
```diff
@@ -556,30 +603,34 @@ impl Session {
...
+                    .await
+                    .context("error initializing persistent DHT")?
+                } else {
+                    let listen_addr = dht_listen_addr(dht_config.port, None, opts.ipv4_only);
                     DhtBuilder::with_config(DhtConfig {
-                        bootstrap_addrs: opts.dht_bootstrap_addrs.clone(),
+                        bootstrap_addrs: dht_config.bootstrap_addrs,
                         cancellation_token: Some(token.child_token()),
                         bind_device: bind_device.as_ref(),
+                        listen_addr: Some(listen_addr),
```
chantra, line 624: hum, I don't think `dht_listen_addr` should leak here. I should move this within DhtBuilder probably, let me take another look.

chantra: actually, this is prior art. The plus side is that it would now obey ipv4_only.

crates/dht/src/persistence.rs
ikatson: I think the only possible values are Ipv*Addr::UNSPECIFIED.

crates/rqbit/src/main.rs
```diff
@@ -603,14 +604,28 @@ async fn async_main(mut opts: Opts, cancel: CancellationToken) -> anyhow::Result
         ..Default::default()
     });
 
+    let dht_bootstrap_addrs = opts
```
ikatson: would be nice to move this block under "else" so that it doesn't do anything if dht is disabled

ikatson (approved):

crates/dht/src/persistence.rs
chantra: ok, this is now handled in `dht_listen_addr` based off of `ipv4_only` value.

crates/rqbit/src/main.rs
chantra: done, this is the only change in the latest force-pushed version https://github.com/ikatson/rqbit/compare/46cf7b17eb5126010a47e7e5ea38da875d7acf13..a476b3cac94923153ca8d8ea0c1bcc93810846bc

chantra (commented):
ok. addressed your comment. LMK if there is anything that still stands out.
Happy to rebase if needed.

ikatson (approved):

ikatson:
LGTM, thanks for the contribution!

Issue #525, opened 2026-01-08
Memory leak / FD exhaustion


0xAcid:
Hi,
I have been using librqbit for almost a year without issue. I recently changed the trackers and I am facing issues since then. It's difficult so far to pinpoint the real issue, here is what I got so far:
I wrapped librqbit into an long-lived app (server-like). After I switched trackers, I have been hitting a file descriptor limit (my default was set pretty low). I increased the limit, but I am now facing out of memory issues after a while that seems to be coming from librqbit.
I traced allocations with `dhat` and what seems to be the culprit is that I keep accepting connections from peers, even if the torrent is completed, and I'm supposed to be in no_upload mode.
- This is the dhat tracking results for block allocations: https://pastebin.com/sc0PVXPA
- This is the dhat tracking results for size allocations: https://pastebin.com/CYr85wNw

On the side of my app, the way I am using librqbit is that I start a torrent, and leave it available for streaming. I never "pause" or "stop" the torrent so I guess it is always running. But I am not sure why I am still seeing peers despite having the torrent completed and upload disabled?
I am experimenting with setting timeouts on `session.peer_opts`, but it is too early to tell if this fixes the issue.

I am not sure this is a librqbit issue, and maybe more the way I am using the library, but any insight/help would be greatly appreciated.
EDIT: I'm using v8.1.1

0xAcid:
So I tried a patch (v8.1.1 so code does not exactly match master):

``` diff
diff --git a/crates/librqbit/src/session.rs b/crates/librqbit/src/session.rs
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -821,6 +821,10 @@
                     }
                 },
                 Some(Ok((live, checked))) = futs.next(), if !futs.is_empty() => {
+                    let session = session.upgrade().context("session is dead")?;
+                    if live.is_finished() && session._disable_upload {
+                        return Ok(())
+                    }
                     if let Err(e) = live.add_incoming_peer(checked) {
                         warn!("error handing over incoming connection: {e:#}");
                     }
```
Which seems to fix the issue. I'm not accepting any more peers if the download is completed and upload is disabled. Though, this still raises the question of using the library with upload enabled. Maybe a limit of simultaneous connections? (Maybe this exists already actually, I don't know the codebase well yet).

ikatson:
Hi, thanks for the report,

[...]

PR #592, opened 2026-05-21
Limit pending incoming handshake checks

puneetdixit200:
## Summary
- cap pending incoming handshake checks before accepting another socket
- let existing handshake validation futures drain instead of allowing burst traffic to keep growing the socket backlog
- add unit coverage for the cap boundary

## Rationale
Addresses #525. The reported FD exhaustion path can happen when incoming connection bursts create many pending handshake checks, each holding a socket until validation finishes or times out. Bounding the pending checks keeps the listener from accepting endlessly during a burst while normal accept behavior remains unchanged below the cap.

## Validation
- RED: `cargo test -p librqbit test_incoming_handshake_backlog_cap -- --nocapture` failed before implementation because the backlog helper/cap did not exist
- `cargo test -p librqbit test_incoming_handshake_backlog_cap -- --nocapture`
- `cargo fmt --all -- --check`
- `cargo check -p librqbit`
- `cargo test -p librqbit session::tests::test_torrent_file_from_info_and_bytes`
- `cargo test -p librqbit session::tests::test_incoming_handshake_backlog_cap`
- `cargo test -p librqbit --lib`
- `cargo clippy -p librqbit --all-targets` (completed; existing `librqbit-core` collapsible-match warning outside this change)
- `cargo test -p librqbit`
- `git diff --check -- crates/librqbit/src/session.rs`

## AI Assistance Disclosure
This contribution was prepared with OpenAI GPT-5.5 via Codex. I reviewed the issue discussion, repository policy, current listener code, and all generated code/test changes before submission.

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -76,6 +76,12 @@ pub const SUPPORTED_SCHEMES: [&str; 3] = ["http:", "https:", "magnet:"];
 
 pub type TorrentId = usize;
 
+const MAX_PENDING_INCOMING_HANDSHAKES: usize = 64;
+
+fn should_accept_more_incoming_handshake_checks(pending_handshakes: usize) -> bool {
+    pending_handshakes < MAX_PENDING_INCOMING_HANDSHAKES
+}
+
 struct ParsedTorrentFile {
     meta: TorrentMetaV1Owned,
     torrent_bytes: Bytes,
@@ -913,7 +919,7 @@ impl Session {
 
         loop {
             tokio::select! {
-                r = l.accept() => {
+                r = l.accept(), if should_accept_more_incoming_handshake_checks(futs.len()) => {
                     match r {
                         Ok((addr, (read, write))) => {
                             trace!("accepted connection from {addr}");
@@ -1745,7 +1751,10 @@ mod tests {
     use itertools::Itertools;
     use librqbit_core::torrent_metainfo::{TorrentMetaV1, torrent_from_bytes};
 
-    use super::torrent_file_from_info_bytes;
+    use super::{
+        MAX_PENDING_INCOMING_HANDSHAKES, should_accept_more_incoming_handshake_checks,
+        torrent_file_from_info_bytes,
+    };
 
     #[test]
     fn test_torrent_file_from_info_and_bytes() {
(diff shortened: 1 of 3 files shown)


crates/librqbit/src/session.rs
```diff
@@ -76,6 +76,12 @@ pub const SUPPORTED_SCHEMES: [&str; 3] = ["http:", "https:", "magnet:"];
 
 pub type TorrentId = usize;
 
+const MAX_PENDING_INCOMING_HANDSHAKES: usize = 64;
+
+fn should_accept_more_incoming_handshake_checks(pending_handshakes: usize) -> bool {
```
ikatson: This can be inlined into select

crates/librqbit/src/session.rs
```diff
@@ -76,6 +76,12 @@ pub const SUPPORTED_SCHEMES: [&str; 3] = ["http:", "https:", "magnet:"];
 
 pub type TorrentId = usize;
 
+const MAX_PENDING_INCOMING_HANDSHAKES: usize = 64;
```
ikatson: This should be in listeneropts

crates/librqbit/src/session.rs
```diff
@@ -1768,4 +1777,18 @@ mod tests {
         assert_eq!(parsed.info, generated_parsed.info);
         assert_eq!(parsed_trackers, get_trackers(&generated_parsed));
     }
+
```
ikatson: No need to test a comparison

ikatson (changes requested):

puneetdixit200:
Addressed in c750b46:

- the accept guard is now inlined in the `select!`
- the cap lives in `ListenerOptions` with a default value
- the comparison-only unit test was removed

I also updated the example option construction to use `..Default::default()` so adding the listener option does not force unrelated examples to spell out every field.

puneetdixit200:
Addressed the review follow-up and fixed a branch-specific uTP e2e regression from the lower default cap. I changed the default pending incoming handshake-check cap to 256, matching the reporter's working value while keeping the cap configurable through ListenerOptions and inline in select.\n\nVerified locally:\n- cargo fmt --all -- --check\n- E2E_TIMEOUT=240 cargo test -p librqbit --lib tests::e2e::test_e2e_download_utp -- --exact\n- cargo test -p librqbit --lib\n- cargo check -p librqbit\n- cargo clippy --all-targets (passes; existing librqbit-core collapsible_match warning remains)

PR #605, opened 2026-06-04
Allow pausing during the initial file check

exbit:
Make the initial hash check abortable. pause() now requests a pause when the torrent is still Initializing instead of bailing out; initial_check periodically checks the flag and stops early; and the start path guards against concurrent checks via try_start_check/finish_check and reports the torrent as Paused while a pause is pending.

### DIFF
--- a/crates/librqbit/src/torrent_state/mod.rs
+++ b/crates/librqbit/src/torrent_state/mod.rs
@@ -517,7 +517,7 @@ impl ManagedTorrent {
             let g = self.locked.read();
             match &g.state {
                 ManagedTorrentState::Initializing(i) => {
-                    resp.state = if g.paused { S::Paused } else { S::Initializing };
+                    resp.state = S::Initializing { paused: g.paused };
                     resp.progress_bytes = i.checked_bytes.load(Ordering::Relaxed);
                 }
                 ManagedTorrentState::Paused(p) => {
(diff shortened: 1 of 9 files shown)


crates/librqbit/src/torrent_state/mod.rs
```diff
@@ -495,10 +513,11 @@ impl ManagedTorrent {
...
         };
 
-        self.with_state(|s| {
-            match s {
+        {
+            let g = self.locked.read();
+            match &g.state {
                 ManagedTorrentState::Initializing(i) => {
-                    resp.state = S::Initializing;
+                    resp.state = if g.paused { S::Paused } else { S::Initializing };
```
ikatson: On mobile so can't test how it looks, but I imagine it'll show up as paused as if progress bytes are actually downloaded bytes.

Might make sense to introduce s::initializing{paused: bool}

So that in the UI and API it's reflected properly.
And of course update webui accordingly

ikatson (changes requested):

crates/librqbit/src/torrent_state/mod.rs
exbit: The progress bar will display the number of bytes verified. I've taken your feedback into account and added this to the webui. 

image

ikatson:
Missed your comment previously, thanks for the ping and for the PR

PR #606, opened 2026-06-04
Expose peer client name in peer stats

exbit:
Parse the client name/version from the BitTorrent extended handshake (the "v" field) and store it on the live peer state, surfacing it through PeerStats so the client of each peer can be displayed.


ikatson:
Lgtm, is it possible to plug that into webui peers table too?

exbit:
> Lgtm, is it possible to plug that into webui peers table too?

@ikatson Added

PR #610, opened 2026-06-18
Add ManagedTorrent::output_folder() accessor

### LINKED ISSUE #609: Expose a public accessor for a torrent's resolved output folder on ManagedTorrent
### Problem

When embedding `librqbit` as a library, there's no public way to learn where a torrent's data actually lives on disk. `ManagedTorrent::shared().options` is `pub(crate)`, and the resolved `output_folder` lives inside `ManagedTorrentOptions`, so external code can't read it. (Still the case in `9.0.0-rc.0`.)

The session-wide default is available (`Session::new(default_output_folder)`), but the *per-torrent* resolved folder isn't — and they differ whenever a torrent is added with an `output_folder`/`sub_folder` override, or gets the auto-subfolder for a multi-file torrent.

### Workaround (and why it's fragile)

I reconstruct the path from the session's default download dir plus the multi-file subfolder rule (`name` for ≥2 files, none for single-file). This breaks for any per-torrent `output_folder` override and duplicates logic that already lives in `Session`.

### Request

A public accessor, e.g.:

```rust
impl ManagedTorrent {
    /// The resolved on-disk folder this torrent's files are written under.
    pub fn output_folder(&self) -> &Path {
        &self.shared().options.output_folder
    }
}
```

A per-file absolute-path helper (joining `output_folder` with each `FileInfo::relative_filename`) would be an even nicer convenience, but the folder accessor alone unblocks the use case (a UI that opens/reveals downloaded files).

zemse:
`shared()` exposes most of `ManagedTorrentShared`, but `output_folder` lives in its `pub(crate)` `options` field (and `ManagedTorrentOptions` isn't public), so embedders can't reach it from outside the crate. This adds a small getter returning `&Path`.

Lets embedders learn where a torrent's files are written (e.g. to open/reveal them).

Closes #609.

### DIFF
--- a/crates/librqbit/src/torrent_state/mod.rs
+++ b/crates/librqbit/src/torrent_state/mod.rs
@@ -7,7 +7,7 @@ pub mod utils;
 
 use std::collections::HashSet;
 use std::net::SocketAddr;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::Weak;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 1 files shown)


ikatson:
Thanks for your contribution!

PR #614, opened 2026-06-28
Support WANIPConnection v2 for UPnP port forwarding

### LINKED ISSUE #612: UPnP IGD (port forwarding) v2.0 support
Currently port forwarding doesn't work if you have a upnp-igd v2.0 router. OpenWRT's miniupnpd for instance defaults to v2 protocol only. It would be excellent if we had v2 support as well.

rektide:
The service-type filter matched only WANIPConnection:1 exactly, but modern routers commonly advertise WANIPConnection:2 (e.g. some OpenWrt builds). This caused get_wan_ip_control_urls to return zero matches and silently skip forwarding, even though miniupnpc worked against the same IGD.

Match either v1 or v2, and thread the matched service-type URN through to the SOAP AddPortMapping body and SOAPAction header (previously hardcoded to :1).

Resolves #612

### DIFF
--- a/crates/upnp/src/lib.rs
+++ b/crates/upnp/src/lib.rs
@@ -14,10 +14,19 @@ use tokio::sync::mpsc::{UnboundedSender, unbounded_channel};
 use tracing::{Instrument, Span, debug, debug_span, trace, warn};
 use url::Url;
 
-const SERVICE_TYPE_WAN_IP_CONNECTION: &str = "urn:schemas-upnp-org:service:WANIPConnection:1";
+const SERVICE_TYPE_WAN_IP_CONNECTION_V1: &str = "urn:schemas-upnp-org:service:WANIPConnection:1";
+const SERVICE_TYPE_WAN_IP_CONNECTION_V2: &str = "urn:schemas-upnp-org:service:WANIPConnection:2";
+
+fn is_wan_ip_connection_service(service_type: &str) -> bool {
+    service_type == SERVICE_TYPE_WAN_IP_CONNECTION_V1
+        || service_type == SERVICE_TYPE_WAN_IP_CONNECTION_V2
+}
+
 const SSDP_MULTICAST_IP: SocketAddr =
     SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(239, 255, 255, 250), 1900));
 pub const SSDP_SEARCH_WAN_IPCONNECTION_ST: &str = "urn:schemas-upnp-org:service:WANIPConnection:1";
+pub const SSDP_SEARCH_WAN_IPCONNECTION_V2_ST: &str =
+    "urn:schemas-upnp-org:service:WANIPConnection:2";
 pub const SSDP_SEARCH_ROOT_ST: &str = "upnp:rootdevice";
 
 pub fn make_ssdp_search_request(kind: &str) -> String {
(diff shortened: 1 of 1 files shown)


crates/upnp/src/lib.rs
```diff
@@ -14,7 +14,14 @@ use tokio::sync::mpsc::{UnboundedSender, unbounded_channel};
...
+const SERVICE_TYPE_WAN_IP_CONNECTION_V2: &str = "urn:schemas-upnp-org:service:WANIPConnection:2";
+
+fn is_wan_ip_connection_service(service_type: &str) -> bool {
+    service_type == SERVICE_TYPE_WAN_IP_CONNECTION_V1
+        || service_type == SERVICE_TYPE_WAN_IP_CONNECTION_V2
+}
+
 const SSDP_MULTICAST_IP: SocketAddr =
     SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(239, 255, 255, 250), 1900));
 pub const SSDP_SEARCH_WAN_IPCONNECTION_ST: &str = "urn:schemas-upnp-org:service:WANIPConnection:1";
```
ikatson, line 27: It seems only v1 is discovered, this probably needs to be updated too

ikatson (changes requested):
This looks great, thanks for the PR. Please address the comment otherwise lgtm

crates/librqbit/src/listen.rs
```diff
@@ -16,6 +16,7 @@ pub(crate) struct ListenResult {
     pub tcp_socket: Option<TcpListener>,
     pub utp_socket: Option<Arc<UtpSocketUdp>>,
     pub enable_upnp_port_forwarding: bool,
+    pub disable_upnp_port_forward_v2: bool,
```
ikatson: This flag is excessive, let's keep it simple

crates/librqbit/src/listen.rs
```diff
@@ -15,7 +15,7 @@ use crate::{stream_connect::ConnectionKind, vectored_traits::AsyncReadVectored};
 pub(crate) struct ListenResult {
     pub tcp_socket: Option<TcpListener>,
     pub utp_socket: Option<Arc<UtpSocketUdp>>,
-    pub enable_upnp_port_forwarding: bool,
+    pub upnp_search_kinds: Vec<String>,
```
ikatson: This is a breaking change, why change any config options at all? Why not just unconditionally support both like the earlier change? 
I don't see why users would care.
If there's a good reason to keep configurability, I'd keep the old flag and add an optional v1/v2 both parameter.

ikatson (changes requested):

rektide:
@ikatson thanks for the very good review, my apologies for not getting it right.

i replaced the old branch with a new one that i believe better satisfies! no rush on review. thanks for the wonderful project.

crates/upnp/src/lib.rs
rektide: :facepalm: oh man. thank you for catching this. bah!

ikatson:
What happens when the same endpoint advertises both v1 and v2? Does it happen in practice?
Do we explicitly forward to both of them like they are independent and hope for the endpoint to figure this out? It's fine if so, just want to make sure.

Otherwise LGTM, if you ensured it works. I need to check on my router though also, to make sure it didn't regress

PR #616, opened 2026-06-28
Fix DefaultBodyLimit to allow torrents bigger than 2 MB to be uploaded, Fixed init state throwing 500 internal server error

libalpm64:
- When a torrent is in the Initialized or Error state we should return an empty zeroed bitfield instead of throwing and error as a 500 internal server error.
- Raised body limit size when posting from the axum default of 2 MB to 100 MB. Fixes the 413 payload to large error when uploading huge torrent files.

### DIFF
--- a/crates/librqbit/src/api.rs
+++ b/crates/librqbit/src/api.rs
@@ -500,22 +500,15 @@ impl Api {
 
     pub fn api_dump_haves(&self, idx: TorrentIdOrHash) -> Result<(BF, u32)> {
         let mgr = self.mgr_handle(idx)?;
-        let result = mgr.with_chunk_tracker(|chunks| {
+        mgr.with_chunk_tracker(|chunks| {
             let bf = BF::from_bitslice(chunks.get_have_pieces().as_slice());
             let len = chunks.get_lengths().total_pieces();
             (bf, len)
-        });
-        match result {
-            Ok(r) => Ok(r),
-            Err(_) => {
-                let len = mgr
-                    .with_metadata(|m| m.lengths().total_pieces())
-                    .unwrap_or(0);
-                let bf =
-                    BF::from_boxed_slice(vec![0u8; (len as usize).div_ceil(8)].into_boxed_slice());
-                Ok((bf, len))
-            }
-        }
+        })
+        .with_status_error(
+            StatusCode::PRECONDITION_FAILED,
+            crate::Error::TorrentIsNotLive,
+        )
     }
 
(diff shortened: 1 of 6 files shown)


crates/librqbit/src/http_api/handlers/torrents.rs
```diff
@@ -35,11 +36,14 @@ pub async fn h_torrents_post(
     State(state): State<ApiState>,
     Query(params): Query<TorrentAddQueryParams>,
     Timeout(timeout): Timeout<600_000, 3_600_000>,
-    data: Bytes,
+    body: Body,
 ) -> Result<impl IntoResponse> {
     let is_url = params.is_url;
     let opts = params.into_add_torrent_options();
-    let data = data.to_vec();
+    let data = to_bytes(body, 100 * 1024 * 1024)
```
ikatson: Can we make this configurable? Ie expose in httpapiListenOpts

Also the default seems excessive, how about 10mb?

crates/librqbit/src/api.rs
```diff
@@ -500,11 +500,22 @@ impl Api {
...
             (bf, len)
-        })?)
+        });
+        match result {
+            Ok(r) => Ok(r),
+            Err(_) => {
+                let len = mgr
+                    .with_metadata(|m| m.lengths().total_pieces())
+                    .unwrap_or(0);
+                let bf =
```
ikatson: Why zeroes instead of a better http error code? 

500 internal server error isn't great for sure, but allocating an empty bit vector + "lying" about its content isn't great either

ikatson (changes requested):

crates/librqbit/src/http_api/handlers/torrents.rs
libalpm64: I did think of this, but I don't see exactly why the limit matters; it's only going to be used by the webgui/desktop. I will make it 10 MB as a preference (since you realistically are not going to go over 10 MB). Also, regarding configurability, that does sound like a good idea to me.

crates/librqbit/src/api.rs
libalpm64: > Why zeroes instead of a better http error code?
> 
> 500 internal server error isn't great for sure, but allocating an empty bit vector + "lying" about its content isn't great either

Yeah, I think this is probably the wrong approach, thinking back at this, the main goal I was trying to achieve was to stop the UI from erroring during the normal init to the pause window, rather than fixing the reporting itself.

ikatson:
Thanks for this, looks great

PR #621, opened 2026-07-05
feat: release file handles when a torrent is paused or stopped

exbit:
Adds a release_files() storage hook that closes OS file handles on pause (including a pause during the initial check), so files can be moved/updated/deleted while the torrent stays registered. Pausing an already-paused torrent now re-releases handles instead of failing.


exbit:
@ikatson Hi, could you take a look?

ikatson:
Hey @exbit, the code and change look good, I want to check manually before merging. Will merge once I confirm 

Sorry for delay!

Issue #425, opened 2025-06-21
RFE: implement a UPNP/DLNA controller


gdamjan:
> A DLNA Controller, also known as a Digital Media Controller (DMC), is a device that can find and control media on a DLNA server and send it to a DLNA renderer for playback. Essentially, it acts as a remote control, allowing you to browse and select content from a server and direct it to another device for display or playback.

This would allow to click "Play" from the rqbit interface (web or desktop) and get it played on your TV. Some inspiration can be seen in [Go2tv](https://github.com/alexballas/go2tv/) or the rust library/cli [crab-dlna](https://github.com/gabrielmagno/crab-dlna).

Some nice features on top of this would also be:
- transcoding with ffmpeg (go2tv does that)
- subtitle control and downloading

gdamjan:
Just added this RFE (request for enhancement) to start the discussion. I can try implementing this if it's considered "in scope" for rqbit

ikatson:
> I can try implementing this if it's considered "in scope"

Whatever enhances the builtin Upnp/DLNA to make it friendlier can definitely be in scope. And if you want to contribute, I'll be glad to help.

> transcoding

I've been missing transcoding a bit as some files (especially some audio streams) can't be played on my TVs. So I had an idea to try to do it, this would be much more involved though as you can't seek with byte ranges any more, and you'll need to seek with timestamps and pass it along to ffmpeg (which in turn can use the http:// current stream implementation as input)

Anyway, if you're eager to try out doing this, I encourouge you try :)

PR #623, opened 2026-07-15
feat: introduce an upnp-control-point crate

gdamjan:
- maintains a hashmap of upnp media renderers (ie. TVs) keeping a unique id, a friendly name, their location url and optional rendering control url and av transport control url.

- will run a background task (an endless loop) to find renderers and purge stale items.

- implements play, pause, stop, seek, set_volume and get_status for the renderers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2901,6 +2901,7 @@ dependencies = [
  "librqbit-sha1-wrapper",
  "librqbit-tracker-comms",
  "librqbit-upnp",
+ "librqbit-upnp-control-point",
  "librqbit-upnp-serve",
  "librqbit-utp",
  "lru",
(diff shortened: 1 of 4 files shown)


gdamjan:
a later PR¹ will expose this over the http api, and then in the end allow the frontend to control TVs and play from the web-ui to a TV directly.


¹ or I can do multiple commits to this same PR?

ikatson:
Can you describe how it's going to be used? Why would I control my TV from rqbit web page?

gdamjan:
> Can you describe how it's going to be used? 

sure. this is me actually getting back to an old idea reported here #425

> Why would I control my TV from rqbit web page?

in short, while rqbit already is an upnp-server, which can be used by most TVs as-is, The truth is most TV upnp browsers are subpar, slow and clunky.

The rqbit web-ui has much more approachable interface, and you can filter, search, sort movies, and if a person already has it open, it'd be great to also be able to just click a button and for the media to start playing.

At least that's my motivation. If it doesn't make sense for rqbit, that's fine too :)

PR #624, opened 2026-07-22
categories: init

### LINKED ISSUE #267: feature request: support tags and get your app added to *arr apps and flexget
Hi,

This looks really exciting, thank you so much for creating this. It would be awesome if I could use this as a downloader in the *arr apps (sonarr, radarr, lidarr, readarr, etc) and flexget. To that end, the only obviously missing feature is tags to support specific download endpoints per program (speed limits, etc. also appear to be missing but those are nice to have vs dealbreaker features) and to get this app added as an option in those apps. Any chance this is possible in the near term?

Thanks again.

MarcelCoding:
closes #267

- each torrent can be associated with a category
- categories must be defined using the cli or env vars
- in the webui categories can be assigned when adding torrent or on a later stage in the details view
- the torrent list can be filtered by categories
- the api also supports filtering by categories, so external application can filter for only their torrents

disclamer: this was implemented using ai

### DIFF
--- a/crates/librqbit/src/api.rs
+++ b/crates/librqbit/src/api.rs
@@ -173,6 +173,9 @@ impl TorrentIdOrHash {
 pub struct ApiTorrentListOpts {
     #[serde(default)]
     pub with_stats: bool,
+    /// Filter by categories. Empty means no filter.
+    #[serde(default)]
+    pub categories: Option<String>,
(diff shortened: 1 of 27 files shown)


SuperSandro2000:
@ikatson Could you take a look? Thanks!

Issue #341, opened 2025-03-12
[question] naming


JeromeSchmied:
hi, 
first of all, huge thanks for this app, I find it truly amazing!
The only thing that isn't quite clear is the naming.
Could you tell me where does it come from and what it hides?

ikatson:
Is that a real question? Can you elaborate?

JeromeSchmied:
Yes, it's quite real actually. 
Thing is, I'm afraid the name has some remainders from qBittorrent in it, namely the `qbit` part, prefixed with an `r` that stands for Rust.
If this is the case, I have some further questions as well.
nothing serious though.

ikatson:
That's hilarious 😂 Indeed it shares 4 common letters with qbittorrent!

So what else do you want to know?

ikatson:
> what else does it hide

What can the name hide? The repo is all open, what is there to hide, I don't understand

JeromeSchmied:
I just wanted to know whether the name stands for (hides) `Rust-qBittorrent`, in that case I'd suggest changing it, as the `q` in `qBittorrent` stands for the gui framework that it's written in: [Qt](https://www.qt.io/product/framework), and as this app (afaik) doesn't have anything to do with `Qt`, it would be an odd name. that's all

ikatson:
Let's say it's for quantum qubit :) Closing as irrelevant

JeromeSchmied:
fair enough, I'll try to keep that in mind

PR #633, opened 2026-08-19
feat(mse): Message Stream Encryption (MSE) support

### LINKED ISSUE #617: please add rc4 encryption when downloading
My ISP is now throttling Bittorrent downloads, so all BT clients have to use RC4 to connect to other peers. I hope rqbt can add an RC4 option.

lingdiansr:
Closes #617

## Summary

Adds MSE (Azureus protocol encryption) support to rqbit: DH-768 key exchange, RC4 stream cipher, SHA-1 key derivation, the incoming (responder) and outgoing (initiator) handshakes, plaintext sniffing/fallback, and a three-state `MseMode` config. The handshake lives in `stream_connect.rs` and is consumed by `peer_connection.rs` (outgoing) and `session.rs` (incoming). No dead code: every piece is referenced by a live call site or a protocol test.

Size: 17 files, +1941 / −32.

## What's in it

**`MseMode` config** (`mse/mod.rs`, default `Disabled`) — `Disabled` (plaintext only), `Enabled` (prefer MSE, plaintext redial on failure), `Forced` (require MSE, drop peers that don't complete it). Threaded through `SessionOptions` / `PeerConnectionOptions` / `ManagedTorrentOptions`. The binary gains `--mse-mode disabled|enabled|forced` (env `RQBIT_MSE_MODE`).

**Handshake abstraction** — outgoing: `StreamConnector::connect_with_handshake` → `OutgoingHandshake`; incoming: `accept_with_handshake` → `IncomingHandshake` (a struct, not a tuple: parsed handshake, streams, read buffer, `encrypted` flag). `session.rs` carries no MSE protocol details; its only input is a lazy `Fn() -> Vec<[u8; 20]>` info-hash enumerator, which the acceptor calls only if an initiator actually reaches the SKEY step.

**Crypto, delegated to crates**

- RC4 via the `rc4` crate. The PadB/VC sync uses pattern search (compute the encrypted-VC pattern once, scan the raw bytes for it), matching libtorrent's `read_pe_syncvc` / transmission's `read_vc`, so no cipher-state clone is needed (the crate has no `Clone`/seek). `Rc4Writer` encrypts each write exactly once into a reusable `pending` buffer and never re-encrypts on `Pending` or short-write retries.
- DH-768 via `crypto-bigint` (`FixedMontyForm::pow`). Only the modular exponentiation is delegated: the group is MSE-specified and not an OpenSSL standard group, so the prime is hardcoded with the same hex as libtorrent `pe_crypto.cpp` / transmission `peer-mse.cc` / aria2 `MSEHandshake.cc`.
- SHA-1 goes through the existing `sha1w` wrapper.
- On hardware acceleration: OpenSSL 3 removed RC4 and the 768-bit group isn't one of OpenSSL's, so there is no hardware path for either (libtorrent/transmission/aria2 keep both inline for the same reason); `sha1w` does use a hardware backend (aws-lc-rs) when the `rust-tls` feature is enabled.

**Wiring** — MSE applies to any connection kind (TCP/uTP/SOCKS); there is no connection-kind check any more. Plaintext peers are accepted unless `Forced`; an outgoing MSE failure or plaintext detection redials plaintext on a fresh connection. Every path of the incoming handshake read is bounded by `rwtimeout`, including the plaintext path (a peer that sends the 20-byte BT prefix and then goes silent can no longer hold an acceptor slot open indefinitely).

**Tests (18 MSE tests)** — dh768 external bigint vectors; a key-derivation KAT with independently computed constants (keyA/keyB SHA-1 inputs, drop-1024 position, VC pattern); stream-wrapper state machine (Pending / short write / error); duplex handshake; plaintext sniff fallback; zero-length IA; fragmented prefix replay; plaintext-prefix stall timeout; fresh-redial; Disabled single-connection; Forced failure; default-to-`Disabled` lock.

## Default value

`MseMode` defaults to `Disabled`, so this merge is a zero-behavior change: every existing connection takes the plaintext path exactly as before, and the default path does no extra work (the info-hash enumerator is never called, and no per-connection snapshot is built). Users opt in via `SessionOptions::mse_mode` or `--mse-mode`. This follows the precedent of the uTP support PR (#322, "disabled by default"). A unit test locks the default to `Disabled`.

## Provenance

This MSE implementation was developed on my rqbit-9 fork (`lingdiansr/rqbit`, branch `mse-dev`, based on v9.0.0) to fix a real-world problem: librqbit lacked MSE, so peers that require encryption (Xunlei, BitComet, etc.) dropped plaintext handshakes — swarms ran 85–97% dead peers and downloads stalled below 1 MB/s (FluxDown issue #341). The implementation follows the Azureus/libtorrent MSE spec (DH-768 + RC4 + SHA-1 key derivation); the fork is consumed by FluxDown PR 475 via a git dependency, where it measured 29 MB/s (Ubuntu) and 20–24.5 MB/s (Win11 pure-DHT magnet) vs <1 MB/s before, with encrypted peer connections now succeeding.

## Verification

- `cargo check -p librqbit --all-targets` — clean
- `cargo test -p librqbit --lib` — 50 passed, 0 failed, 5 ignored (incl. the 18 MSE tests)
- `cargo clippy -p librqbit --all-targets` — 0 warnings
- `cargo fmt --all -- --check` — clean
- `cargo check -p rqbit --no-default-features --features default-tls,postgres,prometheus --all-targets` — clean (the `webui` feature's build script runs `npm install && npm run build`; npm is not installed in my environment, so the binary is checked with that feature off)
- `cargo check --workspace --all-targets --exclude rqbit --exclude rqbit-desktop` — clean
- `cargo run -p rqbit -- --help` lists `--mse-mode`
- The plaintext-stall regression test fails before the fix (hangs until the 2s outer timeout) and passes after it (~200 ms), i.e. the read is now bounded by `rwtimeout`

## Notes on authorship

The implementation was developed with AI assistance (code generation and review), then verified by hand: I traced the 9.0 peer-I/O adaptation points (`into_vectored_compat`, the single-`read` `ReadBuf` limitation), ran the test suite, and validated the crypto primitives against external test vectors. The rework after your review (crate-based crypto, the handshake abstraction, lazily enumerated info hashes) was also AI-assisted, with each change reviewed and verified item by item against the resulting code.


ikatson:
Thanks for the PR

I will not be merging dead code, we need a change of strategy here - got burnt by this when someone tried doing btv2 recently and never finished- this left useless dead code around. 

Either send a stack of PRs that implement the feature fully, 
Or implement top down, not bottom up, with anyhow::bail stubs for not yet implemented pieces.
Alternatively, if the core needs to be refactored to be flexible enough to support encryption (eg readbuf), those can be separate PRs in the stack. If those refactors are reasonable I might merge as is even if encryption is never committed.

lingdiansr:
> Thanks for the PR
> 
> I will not be merging dead code, we need a change of strategy here - got burnt by this when someone tried doing btv2 recently and never finished- this left useless dead code around.
> 
> Either send a stack of PRs that implement the feature fully, Or implement top down, not bottom up, with anyhow::bail stubs for not yet implemented pieces. Alternatively, if the core needs to be refactored to be flexible enough to support encryption (eg readbuf), those can be separate PRs in the stack. If those refactors are reasonable I might merge as is even if encryption is never committed.

Thanks for the detailed feedback — this is exactly the guidance I needed, and I've restructured the PR accordingly.

**Top-down, implemented fully, no dead code.** Instead of the bottom-up primitives-first approach, I've reorganized the work into a top-down series following the "wiring → handshake" order you described:

1. `feat(mse): add MseMode config and wire MSE into peer connections` — the config surface (`MseMode` three-state, default `Disabled`) threaded through `SessionOptions` / `PeerConnectionOptions` / `ManagedTorrentOptions`, plus the two call sites (`connect_with_mse_fallback` in `peer_connection.rs`, `check_incoming_connection` in `session.rs`) — at this commit they call `bail!` placeholders, exactly as you suggested.
2. `feat(mse): implement incoming MSE handshake (responder side)` — fills in the incoming hook with the real handshake.
3. `feat(mse): implement outgoing MSE handshake with plaintext fallback` — fills in the outgoing hook (including the plaintext redial fallback).

Every commit is a complete, compilable slice, and the final PR has **no dead code** — every function is referenced by a live call site or a protocol test. I also added a unit test locking `MseMode::default() == Disabled`, so the feature merges as a zero-behavior change (following the uTP "disabled by default" precedent).

To be explicit about the lessons from btv2: nothing here is a placeholder left for a later commit — all three commits together form the full MSE implementation (DH-768 + RC4 + SHA-1 key derivation, incoming/outgoing handshakes, plaintext sniffing and fallback), all wired and tested. The top-down ordering is only about commit structure so each step is reviewable independently.

crates/librqbit/src/mse/mod.rs
```diff
@@ -0,0 +1,63 @@
+//! Message Stream Encryption (MSE) support.
+//!
+//! MSE obfuscates the BitTorrent handshake and stream with RC4 and a DH-768
+//! key exchange (Azureus protocol encryption). Some peers (Xunlei, BitComet,
+//! etc.) refuse plaintext handshakes, so MSE is required to talk to them.
+//!
+//! This module is built top-down: the [`MseMode`] config and the peer
```
ikatson: > This module is built top-down ...

nit: this is not worth keeping in the code

crates/librqbit/src/mse/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/// Initiate MSE on a connected stream.
+///
+/// Not yet implemented: this placeholder is wired into the outgoing peer
+/// connection path and will be replaced by the full handshake.
+pub async fn outgoing<R, W>(
+    read: R,
+    write: W,
+    info_hash: &[u8; 20],
```
ikatson: let's use Id20

crates/librqbit/src/error.rs
```diff
@@ -30,6 +32,8 @@ pub enum Error {
     WrongInfoHash,
     #[error("connecting to ourselves")]
     ConnectingToOurselves,
+    #[error("MSE is forced, but peer {0} did not complete the MSE handshake")]
+    MseForced(SocketAddr),
```
ikatson: SocketAddr not needed, let's remove, no other error has the peer addr directly in it

crates/librqbit/src/peer_connection.rs
```diff
@@ -120,6 +126,38 @@ struct ManagePeerArgs {
...
+    info_hash: &[u8; 20],
+    connect_timeout: Duration,
+    _rwtimeout: Duration,
+    initial_payload: &[u8; 68],
+    mse_mode: MseMode,
+) -> Result<(ConnectionKind, BoxAsyncReadVectored, BoxAsyncWrite, bool)> {
+    let (ckind, read, write) =
+        with_timeout("connecting", connect_timeout, connector.connect(addr)).await?;
+
+    if mse_mode != MseMode::Disabled && matches!(ckind, ConnectionKind::Tcp) {
```
ikatson: Why the Tcp check? is it only supported via tcp?

crates/librqbit/src/peer_connection.rs
```diff
@@ -217,27 +255,41 @@ impl<H: PeerConnectionHandler> PeerConnection<H> {
...
-            let hsz = handshake.serialize_unchecked_len(&mut *write_buf);
-            with_timeout(
-                "writing",
-                rwtimeout,
-                write
-                    .write_all(&write_buf[..hsz])
-                    .map_err(Error::WriteHandshake),
-            )
-            .await?;
+            // When MSE succeeded it already consumed the handshake as IA.
```
ikatson: what is IA?

crates/librqbit/src/peer_connection.rs
```diff
@@ -120,6 +126,38 @@ struct ManagePeerArgs {
...
+        with_timeout("connecting", connect_timeout, connector.connect(addr)).await?;
+
+    if mse_mode != MseMode::Disabled && matches!(ckind, ConnectionKind::Tcp) {
+        // MSE outgoing handshake is not yet implemented: the placeholder
+        // consumes the streams and always errors out, which propagates to
+        // the caller. This path is unreachable while the placeholder bails.
+        crate::mse::outgoing(read, write, info_hash, initial_payload)
+            .await
+            .map_err(Error::Anyhow)?;
+        unreachable!("MSE outgoing placeholder never returns Ok");
```
ikatson: why is this line needed?

crates/librqbit/src/mse/rc4.rs
```diff
@@ -0,0 +1,140 @@
+//! RC4 (ARC4) stream cipher - self-contained, zero-dependency implementation
```
ikatson: why not use a crate for it? I haven't studied their applicability, but I know LLMs like to invent the wheel for everything

same goes for DH code

crates/librqbit/src/session.rs
```diff
@@ -929,11 +944,109 @@ impl Session {
             bail!("Incoming ip {incoming_ip} is not in allowlist");
         }
 
-        let mut read_buf = ReadBuf::new();
-        let h = read_buf
-            .read_handshake(&mut reader, rwtimeout)
+        if self.mse_mode == MseMode::Disabled {
```
ikatson: this seems to pollute session code with mse details too much, can we abstract it away somehow?

This is also a bit pervasive in peer_connection too. I'm not sure it can be done better, but want to know your thoughts on how to keep the core logic readable and composable.

Without reading too much on how this works, I expected maybe smth like this:

stream_connector (or some wrapper around it)::connect_with_handshake(peer_id) -> (BoxAsyncRead, BoxAsyncWrite)

and for incoming:

smth.accept() -> (Details, BoxAsyncRead, BoxAsyncWrite), where Details contains everything you need to know like connection kind, if it's encrypted or not etc. Not a tuple, a struct probably, just to show the idea.

crates/librqbit/src/peer_connection.rs
```diff
@@ -120,6 +126,91 @@ struct ManagePeerArgs {
...
+    initial_payload: &[u8; 68],
+    mse_mode: MseMode,
+) -> Result<(ConnectionKind, BoxAsyncReadVectored, BoxAsyncWrite, bool)> {
+    use crate::mse::OutgoingOutcome;
+    use crate::vectored_traits::AsyncReadVectoredIntoCompat;
+
+    let (ckind, read, write) =
+        with_timeout("connecting", connect_timeout, connector.connect(addr)).await?;
+
+    if mse_mode == MseMode::Disabled || !matches!(ckind, ConnectionKind::Tcp) {
```
ikatson: why Tcp only? Why wouldn't uTP or socks support it transparently too?

ikatson (changes requested):
It's not too big, so let's just shrink into one commit, don't bother with splitting into top-down

I haven't finished looking in detail, but this is a first high-level pass.

Main concerns are:
- lots of custom code for crypto. Need to try to use crates for it, esp with hardware acceleration if any (openssl etc)
- spaghetti code where mse becomes intrusive into the rest of the logic

lingdiansr:
> It's not too big, so let's just shrink into one commit, don't bother with splitting into top-down
> Main concerns are:
> - lots of custom code for crypto. Need to try to use crates for it, esp with hardware acceleration if any (openssl etc)
> - spaghetti code where mse becomes intrusive into the rest of the logic

Thanks again for the review — I've reworked the PR accordingly. It's now a single commit, and I've addressed both concerns plus the inline comments. Response below.

**Crypto crates.** You were right about the wheel-invention. To be upfront: I originally started from the libtorrent / transmission implementations (which inline both RC4 and the DH-768 group), and I wasn't familiar enough with the Rust crypto ecosystem to question that assumption — so I mirrored the reference engines rather than looking for a better fit. I've now replaced both:
- **DH-768 → `crypto-bigint`.** The 768-bit group is MSE-specified (not an OpenSSL standard group), so the prime is hardcoded and the modular exponentiation is done by the crate; the hand-rolled bigint `mod_reduce`/`powm` is gone, verified against external vectors.
- **RC4 → `rc4` crate.** Since RustCrypto's `rc4` has no `Clone`/seek, I rewrote the writer following how libtorrent/transmission handle it — the PadB/VC scan computes the encrypted-VC pattern once and searches the raw bytes, and `Rc4Writer` buffers ciphertext and only reports full writes or `Pending`, keeping the keystream position consistent. SHA-1 already goes through `sha1w`.

One caveat on hardware acceleration: RC4 and this fixed DH-768 group have no meaningful HWA available (OpenSSL 3 removed RC4; the group isn't standard), so those two are pure-Rust — only SHA-1 gets a hardware backend today. For a per-connection handshake I don't think that's a real loss.

**Leftover code.** Apologies — the `SocketAddr` on `MseForced` and the `ConnectionKind::Tcp` guard were leftovers from a debugging pass I failed to clean up before opening. Both are gone now: the error carries no address, and MSE applies to any connection kind (uTP/SOCKS included) with `mse_mode` as the only switch.

**Abstraction.** Implemented per your sketch: outgoing via `StreamConnector::connect_with_handshake -> OutgoingHandshake`, incoming via `accept_with_handshake -> IncomingHandshake` (struct, not tuple). `session.rs`/`peer_connection.rs` carry no MSE details — the only thing that stays in the session is building the SKEY→info_hash lookup (it needs the torrent db), passed in as a closure.

Happy to adjust further if anything still looks off.

lingdiansr:
Thanks for the review — I've reworked the branch and pushed. Everything below is in the single commit (`9b7877cb`, rebased on `002a3484`).

**Your inline comments**

1. `error.rs:36` (SocketAddr) — removed; `Error::MseForced` carries no address (`error.rs:34`).
2. `peer_connection.rs:148` / `:151` / `:155` (why Tcp-only / why this line) — that whole guard is gone, including the `debug!` that logged the skip. `connect_with_handshake` (`stream_connect.rs:404`) has no connection-kind check, so MSE runs over TCP, uTP and SOCKS alike.
3. `mse/rc4.rs:1` (why not use a crate?) — done for both: RC4 is the `rc4` crate, DH-768's modexp is `crypto-bigint`. The hand-rolled `rc4.rs` and the bigint `mod_reduce`/`powm` are deleted.
4. `session.rs:947` (MSE details polluting the session) — addressed; see the abstraction note below.

**On the crypto crates and hardware acceleration.** RC4 and DH-768 are both crate-backed now. Two things to flag on the HWA question specifically: OpenSSL 3 removed RC4 outright, and the 768-bit group is MSE-specified rather than one of OpenSSL's standard groups — so neither primitive has a hardware path available (libtorrent/transmission/aria2 keep both inline for the same reason). What is delegated is the modexp, via `crypto-bigint`'s `FixedMontyForm::pow`, with the group hardcoded to the same hex as libtorrent `pe_crypto.cpp` / transmission `peer-mse.cc` / aria2 `MSEHandshake.cc`. SHA-1 still goes through `sha1w`, which does pick a hardware backend (aws-lc-rs) when the `rust-tls` feature is on.

One design point the `rc4` crate forced, since it has no `Clone`/seek: PadB/VC sync can't probe candidate offsets on cloned cipher state the way libtorrent's openssl path does. Instead we compute the encrypted-VC pattern once and search the raw bytes for it — which is what libtorrent's `read_pe_syncvc` and transmission's `read_vc` do anyway — so the decrypt stream lands exactly on the right position with no clone, and `Rc4Writer` only ever reports a write complete once its ciphertext is fully flushed, so a `Pending`/short-write retry never re-advances the keystream.

**On the abstraction.** Implemented along the lines of your sketch: outgoing is `StreamConnector::connect_with_handshake -> OutgoingHandshake`, incoming is `accept_with_handshake -> IncomingHandshake` (a struct carrying the parsed handshake, the streams, the read buffer and an `encrypted` flag). `peer_connection.rs` / `session.rs` now only consume those two results.

The one piece that has to stay session-side is mapping the peer's obfuscated info hash back to a torrent, because only the session has the torrent db. So instead of a precomputed snapshot the session passes a `Fn() -> Vec<[u8; 20]>` closure, and the acceptor calls it only if the initiator actually reaches the SKEY step (`mse/mod.rs:373`, call site at `mse/mod.rs:420`). There is no `if mse_mode == Disabled` left in `session.rs` — that branch existed only because the old code built that snapshot unconditionally, and the snapshot is gone entirely.

If you'd still like this shaped differently (e.g. a trait instead of these two entry points), say the word and I'll rework it.

**Since the last push**

- Rebased onto `main` (`002a3484`) — picks up your `incoming.handshake.supports_extended()` fix.
- Incoming plaintext path: after the plaintext sniff, reading the remaining handshake bytes is now bounded by `rwtimeout`. It was an unbounded `read_exact` — a peer sending just the 20-byte BT prefix and then going silent would hold an acceptor slot forever (the listener caps in-flight handshake checks at 256). Added a regression test that reproduces it: fails before the fix (hangs to the 2s outer timeout), passes after (~200 ms).
- Lazy info-hash enumeration (above).
- Dropped the test-only `mse_applied` field on `OutgoingHandshake` and the two asserts that read it; the fallback tests still assert the observable behaviour (the peer sees the plaintext handshake on the redial, and exactly one connection is consumed).
- Added a key-derivation KAT whose constants were computed outside this repo (Python `hashlib.sha1` + an independent RC4 checked against the public `"Wiki"`/`"pedia"` → `1021BF0420` vector), pinning the keyA/keyB direction, the drop-1024 position and the VC pattern against something other than our own code.
- `#[serde(default)]` on `PeerConnectionOptions::mse_mode` (`peer_connection.rs:89`): it is the only non-`Option` field in that struct, so JSON written before MSE existed (desktop app, HTTP API clients) would otherwise fail to deserialize.
- `--mse-mode disabled|enabled|forced` (env `RQBIT_MSE_MODE`) in the binary, shown in `--help`.
- Doc fixes: the sniff-window comment now states the real cost for a silent plaintext peer, and the stale "built top-down / follow-up commit" module note is gone.
- Hygiene: `.idea/` and the leftover refactor-plan markdown are out of the commit, and `/.idea/` is in `.gitignore`.

Verification on this branch: `cargo test -p librqbit --lib` → 50 passed / 0 failed / 5 ignored (18 MSE tests), `cargo clippy -p librqbit --all-targets` → 0 warnings, `cargo fmt --all -- --check` clean, `cargo check -p librqbit --all-targets` clean. (`cargo check -p rqbit` is run with `--no-default-features --features default-tls,postgres,prometheus` here because the `webui` feature's build script needs npm, which this machine doesn't have.)

PR #636, opened 2026-08-20
fix: bound read_chunk to the requested chunk when serving a piece

### LINKED ISSUE #635: bug: seeding a partially selected torrent fails since 9.0.0 (read_chunk over-reads into absent files)
## Summary

Since 9.0.0, seeding a **partially selected** torrent fails: any piece that does not lie in the final file errors out with `error reading chunk`, the requesting peer marks the piece broken and disconnects, and the transfer stalls.

On a fully downloaded torrent the same code path only causes a harmless over-read, which is likely why this went unnoticed.

## Symptom

Peer requests piece 0 of a two-file torrent where only file 0 was selected and downloaded:

```
manage_incoming_peer{addr=127.0.0.1:55553}:
  error managing peer: error reading chunk: error reading 1 bytes, file_id: 16
```

(Minor separate nit: the two format arguments look transposed at `crates/librqbit/src/file_ops.rs:294` — the message means *file_idx 1, 16 bytes*.)

## Root cause

`FileOps::read_chunk` fills the caller's buffer and only stops when that buffer is exhausted:

```rust
let mut buf = result_buf;          // not truncated to chunk_info.size
for (file_idx, file_info) in self.file_infos.iter().enumerate() {
    ...
    self.files.pread_exact(file_idx, absolute_offset, &mut buf[..to_read_in_file])?;
    buf = &mut buf[to_read_in_file..];
    if buf.is_empty() { break; }
}
```

This function is byte-identical in 8.1.1 and 9.0.0 — **the caller changed.**

8.1.1 sized the buffer to the chunk before reading (`peer_connection.rs`):

```rust
write_buf.resize(PIECE_MESSAGE_DEFAULT_LEN, 0);
let preamble_len = serialize_piece_preamble(&chunk, &mut write_buf);
let full_len = preamble_len + chunk.size as usize;
write_buf.resize(full_len, 0);                   // <-- no longer present in 9.0.0
... read_chunk(&chunk, &mut write_buf[preamble_len..])
```

The 9.x ring-buffer / vectored-I/O rework replaced `write_buf` with a fixed `Box<[u8; MAX_MSG_LEN]>` and dropped that `resize(full_len)`. `read_chunk` now receives a buffer much larger than `chunk.size`, so it keeps walking past the requested chunk into subsequent files, and `pread_exact` fails on the first file that isn't on disk.

Only `full_len` bytes are ever transmitted, so for a complete torrent this is just wasted I/O. For a partially selected one, the unselected files don't exist and the read fails.

As of today `main` still has no `resize`, so I believe this is unfixed after 9.0.0.

## Reproduction

1. Create a two-file torrent.
2. Download with only file 0 selected, so file 1 is never created on disk.
3. Seed it and have a peer request piece 0.
4. The read fails as above.

Confirming it's this and nothing else: writing a small dummy file at file 1's path — the only change — makes the same test pass in ~8s.

## Suggested fix

Slice the buffer to the chunk at the call site:

```rust
read_chunk(&chunk, &mut write_buf[preamble_len..full_len])
```

Happy to open a PR if you'd like.

## Context

Found while upgrading a downstream project from 8.1.1 to 9.0.0 (`default-features = false, features = ["rust-tls"]`). Everything else in the upgrade was clean, and the advisory remediation in #620 is much appreciated — it took our dependency audit from 2 high-severity vulns + 4 unmaintained warnings to zero.

maxiarat1:
Fixes #635.

`full_len` is already computed from `chunk.size` one line above, and returned just below as the write length — but `read_chunk` is handed `&mut write_buf[preamble_len..]`, the whole remaining buffer.

`FileOps::read_chunk` fills whatever buffer it is given, advancing through `file_infos` until the buffer is exhausted. With an oversized buffer it walks past the requested chunk into subsequent files, and `pread_exact` fails on the first file that is not on disk:

```
error managing peer: error reading chunk: error reading 1 bytes, file_id: 16
```

On a complete torrent this is only a wasted read, since only `full_len` bytes are ever transmitted. On a **partially selected** torrent the unselected files do not exist, so serving any piece that is not in the final file fails and seeding cannot work.

8.1.1 avoided this by sizing the buffer before reading (`write_buf.resize(full_len, 0)`); the ring-buffer/vectored-I/O rework in 9.0.0 replaced `write_buf` with a fixed `Box<[u8; MAX_MSG_LEN]>` and the resize went away with it. This restores the same bound at the call site instead.

Verified `cargo check -p librqbit` is clean. Confirming the behaviour downstream: creating the absent file as a stub makes a previously failing selective-seed test pass, and this change fixes it without the stub.

Happy to adjust the approach if you'd prefer the bound applied inside `read_chunk` instead.

### DIFF
--- a/crates/librqbit/src/peer_connection.rs
+++ b/crates/librqbit/src/peer_connection.rs
@@ -433,7 +433,7 @@ impl<H: PeerConnectionHandler> PeerConnection<H> {
                             self.spawner
                                 .block_in_place_with_semaphore(|| {
                                     self.handler
-                                        .read_chunk(&chunk, &mut write_buf[preamble_len..])
+                                        .read_chunk(&chunk, &mut write_buf[preamble_len..full_len])
                                 })
                                 .await
                                 .map_err(Error::ReadChunk)?;


ikatson:
Thanks for the fix! I'll publish 9.0.1 with this asap

PR #639, opened 2026-08-25
feat(peers): allow updating connection timeout

veenix:
## Summary

- store the per-torrent peer connection timeout behind an RwLock
- expose getter and setter methods on ManagedTorrent
- read the current timeout when starting each future outbound peer connection
- leave established peer connections untouched

## Motivation

Callers may want a short connection timeout during latency-sensitive startup, then allow slower peers more time if the torrent has not found sufficient throughput. Recreating the managed torrent loses useful state and existing peer connections. A runtime setter permits that adaptive strategy while limiting the change to future connection attempts.

## Testing

- cargo test -p librqbit peer_connect_timeout_can_be_updated_for_future_connections
- cargo check --workspace
- cargo clippy --workspace --all-targets -- -D warnings
- integrated into Remux with a 1.2s fast timeout and a 3s patient-fallback timeout
- 12th Fail exhausted the 15s fast hedge, entered patient probing, started in 36.5s, and played 60s without a stall
- after restarting Remux, the persisted source resumed in 28.6s and played 60s without a stall

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -1348,7 +1348,7 @@ impl Session {
                 storage_factory,
                 options: ManagedTorrentOptions {
                     force_tracker_interval: opts.force_tracker_interval,
-                    peer_connect_timeout: peer_opts.connect_timeout,
+                    peer_connect_timeout: RwLock::new(peer_opts.connect_timeout),
                     peer_read_write_timeout: peer_opts.read_write_timeout,
                     allow_overwrite: opts.overwrite,
                     output_folder,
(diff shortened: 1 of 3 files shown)


ikatson:
> Callers may want a short connection timeout during latency-sensitive startup

Is there a non hypothetical use case? A real world problem that you have encountered that this PR helps solve

veenix:
I have been contributing to https://github.com/lostb1t/remux. It uses rqbit and I'm implementing automatic torrent selection in order to minimize startup time for playback. The change that I'm making is to make this timeout adaptive to balance minimizing startup latency and robustness in finding a good source.

Remux currently uses a fixed 1.2s peer connection timeout. The change that I am making is implementing a 15s fast phase where we use the existing 1.2s timeout, but if after this fast phase, we fail to make a connection, we increase the timeout to 3s for future connection attempts and also broaden the search to more torrent sources.

Without this setter, Remux must either use the slower timeout for every startup or remove and recreate the torrent, losing discovered peers, metadata, and downloaded pieces.

PR #640, opened 2026-08-25
feat(streaming): make lookahead configurable per stream

veenix:
## Summary

- add a public FileStreamOptions type with configurable lookahead_bytes
- preserve the existing 32 MiB behavior for callers using stream() or api_stream()
- add an optional lookahead_bytes query parameter to the HTTP streaming endpoint
- reject a zero-byte lookahead as an invalid request

## Motivation

Streaming clients sometimes issue small, concurrent range probes before choosing a source. Applying the fixed 32 MiB priority window to every probe can prioritize substantially more data than the caller needs and create avoidable contention. This lets each stream request bound its own lookahead while retaining the current default for existing users.

## Testing

- cargo test -p librqbit stream_queue_uses_the_configured_lookahead
- cargo check --workspace
- cargo clippy --workspace --all-targets -- -D warnings
- integrated into Remux against rqbit 9 and exercised through its browser playback flow
- uncached 1080p playback: The Shawshank Redemption started in 20.2s and played 60s without a stall
- uncached patient-fallback playback: 12th Fail started in 36.5s and played 60s without a stall
- Remux torrent proxy tests passed, including bounded finite-range forwarding

### DIFF
--- a/crates/librqbit/src/api.rs
+++ b/crates/librqbit/src/api.rs
@@ -512,8 +512,18 @@ impl Api {
     }
 
     pub async fn api_stream(&self, idx: TorrentIdOrHash, file_id: usize) -> Result<FileStream> {
+        self.api_stream_with_options(idx, file_id, Default::default())
+            .await
+    }
+
+    pub async fn api_stream_with_options(
+        &self,
+        idx: TorrentIdOrHash,
+        file_id: usize,
+        options: crate::FileStreamOptions,
+    ) -> Result<FileStream> {
         let mgr = self.mgr_handle(idx)?;
-        Ok(mgr.stream(file_id).await?)
+        Ok(mgr.stream_with_options(file_id, options).await?)
     }
 }
 
--- a/crates/librqbit/src/http_api/handlers/streaming.rs
+++ b/crates/librqbit/src/http_api/handlers/streaming.rs
@@ -2,7 +2,7 @@ use std::{io::SeekFrom, sync::Arc};
 
 use anyhow::Context;
 use axum::{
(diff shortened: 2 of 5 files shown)


ikatson:
What's the real world use case for this? What client will use that, and what exactly will it help with for that client?

veenix:
I'm working on Remux (https://github.com/lostb1t/remux) and am implementing an automatic torrent selector. It performs small range reads, which is currently 2.5MiB during the phase phase and 4MiB during the patient probing, across multiple torrent candidates to measure real throughput before choosing one.

With the fixed 32 MiB lookahead, each small probe prioritizes substantially more data than it intends to consume. With 16 concurrent candidates, that can create up to 512 MiB of unnecessary priority windows competing for network, peers, and disk activity.

Remux uses this option to match rqbit’s lookahead to the requested range. Existing callers retain the current 32 MiB default. In browser integration tests, uncached sources started successfully and played for 60 seconds without stalls, including a difficult source that required patient fallback.

ikatson:
@veenix LGTM, I can merge it as-is.

Do you know by any chance why @zond submitted identical PR in https://github.com/ikatson/rqbit/pull/647?

zond:
> @veenix LGTM, I can merge it as-is.
> 
> Do you know by any chance why @zond submitted identical PR in #647?

This is me not checking my agents work enough - it found the need for this feature, and when I asked it to implement it in my fork it also found this PR, and decided this PR already did it the best way.

Then when I wanted to contribute it upstream I didn't check if there were already identical PRs.

Apologies, it's all on me.

PR #644, opened 2026-08-28
Optionally expose per-peer bitfields in PeerStats

### LINKED ISSUE #643: Feature request: per-peer piece availability in `PeerStats`
## What I am trying to do

I am embedding `librqbit` in a desktop application and would like to answer,
for a running torrent: **will this finish?**

That turns out to be a different question from "is it downloading". A torrent
connected to forty peers who all stopped at 6% will not finish; one connected
to six peers who between them hold every piece will. Distinguishing those needs
piece availability — the union of the connected peers' bitfields, or at minimum
a rarest-piece copy count.

## What I found

`PeerStats` (`torrent_state/live/peer/stats/snapshot.rs`) currently carries:

```rust
pub struct PeerStats {
    pub counters: PeerCounters,
    pub state: &'static str,
    pub conn_kind: Option<ConnectionKind>,
    pub client_name: Option<String>,
}
```

`PeerCounters::downloaded_and_checked_pieces` is how many pieces a peer has
sent me, rather than how many that peer holds. `api_dump_haves` returns my own
bitfield. So I do not think either gets me there, though I would be glad to be
wrong about that.

I looked at `main` as well as 9.0.0 and `PeerStats` reads the same on both.

## The underlying data seems to be present

`LivePeerState` (`torrent_state/live/peer/mod.rs`) has, per peer:

```rust
// This is used to track the pieces the peer has.
pub bitfield: BF,
```

which is maintained for piece picking. So the tracking already happens; what I
am missing is a way to read it from outside.

## Things I tried first

I looked for a way to do this without changing the crate:

- `PeerConnectionHandler::on_received_message` receives `Bitfield` and `Have`,
  which would be ideal, but `mod peer_connection` is private so the trait is
  not reachable from a dependent crate.
- `StorageFactory` / `TorrentStorage` are public and injectable through
  `SessionOptions.default_storage_factory`, but they see disk I/O rather than
  peer messages.
- `BitV` / `BitVFactory` are in private modules and relate to persisting my own
  bitfield.
- I could not find an observer or callback field on `SessionOptions`.

If there is an approach I have overlooked, I would much rather use it than ask
for an API change.

## Possible shapes

**A count on `PeerStats`:**

```rust
pub struct PeerStats {
    // ...
    /// How many pieces this peer has, from its bitfield.
    pub have_pieces: u32,
}
```

populated in `impl From<&Peer> for PeerStats`. This would be enough for a
rarest-piece copy count and a swarm-health verdict, which is all I need.

**The bitfield itself**, e.g. as `Option<Vec<u8>>` or behind an accessor. A
larger payload, but it would also allow a per-region availability histogram
rather than a single number. I mention it only because it is the same data; the
count alone would cover my use case.

A third option would be making `PeerConnectionHandler` reachable so embedders
can observe protocol messages directly, though that is a much larger surface to
commit to.

## Offer

If one of these looks reasonable, I am happy to put together a PR — or to adapt
to a different shape if you have one in mind. I did not want to send an
unsolicited patch that changes a public type without checking what you would
prefer first.

Either way, no urgency from my side.

---

Checked against 9.0.0 and `main`. Thanks for the library; the v9 reorganisation
made embedding it noticeably cleaner.

adamgreenwell:
Closes #643.

Three commits. The first two stand alone; the third is the actual ask.

**1. Re-export `PeerStats` and `PeerStatsFilter`.** Both appear in the signature
of the public `per_peer_stats_snapshot` but live in a private module, so a
dependent crate can only get a filter from `Default::default()` and cannot name
either type.

**2. Clear a peer bitfield's padding on ingest.** Your call on the review, and
it is the better one — see below. Independent of the stats work, and takeable
without it.

**3. Add `have_bitfield` and `have_pieces` to `PeerStats`, behind a new
`PeerStatsFilter::include_bitfield`, off by default.**

## Against your three concerns

**Performance for callers that do not need it.** Nothing is computed or copied
unless the flag is set — both fields are `None` through a guard on the same
`match`, so the default path does exactly what it did before. `Default` has the
flag off, which `bitfields_are_off_by_default` pins. The HTTP API and every
existing caller are untouched.

**API complexity.** One new field on the filter and two on `PeerStats`, all
opt-in. I put the count behind the same flag rather than adding a second one:
it is nearly free once the bitfield is in hand, and it saves each caller from
re-deriving it.

**Code complexity.** +155/-18 across three files, and taking your ingest
suggestion made it smaller rather than larger — the counting helper is gone and
`from_peer` no longer needs a `total_pieces` argument. The only structural
change is `From<&Peer> for PeerStats` becoming `PeerStats::from_peer(peer,
include_bitfield)`, since the conversion now takes the flag. `Peer` is
`pub(crate)`, so that impl was not reachable externally.

## Why the bitfield rather than just a count

The issue originally asked for a count; that was my mistake and I corrected it
there. A count gives the *mean* copies per piece, but the useful figure is the
*minimum*. Two peers holding 500 pieces each may overlap entirely or not at
all — identical counts, and only one of those torrents can finish.

## The padding, and the second way in

An earlier revision of this PR sliced to `total_pieces` when counting and left
the stored bitfield alone. You asked for it to be fixed at ingest instead, so
that is what it does now: `on_bitfield` clears the spare trailing bits, and
`count_ones()` means "pieces this peer has" everywhere without each caller
having to know about the padding.

Doing it there surfaced something. `on_have` is a second way in, and it bounds
the index with `get_mut` — which checks against the bitfield's *length*. That
length is byte-padded, so it overshoots the piece count by up to 7, and a peer
could set a padding bit one `Have` message after ingest and undo the masking.
It is bounded by the piece count now, in the same commit. That also means a
peer can no longer claim a piece that does not exist.

Neither was harmful before this, which is presumably why it went unnoticed:
every other reader either slices to `total_pieces` first — as `on_bitfield`'s
own "peer has full torrent" check does — or indexes by an already-validated
piece id. The padding was only ever visible to something that counted the whole
bitfield.

## Why `have_bitfield` is `Vec<u8>` and not `BF`

`PeerStats` derives `Serialize` and `bitvec` is built without its `serde`
feature, so `BitBox` has no `Serialize` impl and cloning the field does not
compile. Enabling the feature would also put `bitvec` in the public API, where
a dependent would need a matching version just to name the type. The allocation
is the same either way.

## Checks

`cargo test -p librqbit` (38 passed, 5 ignored), `cargo fmt --check`, and
`cargo clippy --all-targets -- -D warnings` all clean — the last is stricter
than CI's, and worth running here because `lib.rs` opts into
`clippy::cast_possible_truncation` and the count is a `usize`.

Also running as a `[patch.crates-io]` in a desktop client, on a branch carrying
these three commits plus #645; its own suite passes.

Rebased onto current main. Happy to reshape or drop any of it.

### DIFF
--- a/crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
+++ b/crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
@@ -29,6 +29,31 @@ pub struct PeerStats {
     pub state: &'static str,
     pub conn_kind: Option<ConnectionKind>,
     pub client_name: Option<String>,
+    /// How many pieces this peer has.
+    ///
+    /// `None` unless [`PeerStatsFilter::include_bitfield`] was set and the peer
+    /// is live. Provided alongside the bitfield to save every caller the same
+    /// `count_ones()`.
+    ///
+    /// Distinct from `counters.downloaded_and_checked_pieces`, which is how
+    /// many pieces this peer has sent *us*.
+    pub have_pieces: Option<u32>,
+    /// This peer's bitfield.
+    ///
+    /// `None` unless [`PeerStatsFilter::include_bitfield`] was set and the peer
+    /// is live. The trailing padding is cleared on ingest, so the bits are
+    /// exactly the pieces the peer holds.
+    ///
+    /// Raw bytes rather than the internal `BF`, because this struct is
+    /// `Serialize` and `bitvec` is built without its `serde` feature — and
+    /// naming `BitBox` here would put `bitvec` in the public API, where a
+    /// dependent would need a matching version to read the field.
+    ///
+    /// A count is not a substitute: computing piece availability across a
+    /// swarm — the rarest-piece copy count — needs to know *which* pieces each
+    /// peer holds. Two peers with 500 pieces each may overlap entirely or not
+    /// at all, and only one of those torrents can finish.
+    pub have_bitfield: Option<Vec<u8>>,
 }
 
 impl From<&super::atomic::PeerCountersAtomic> for PeerCounters {
@@ -54,8 +79,9 @@ impl From<&super::atomic::PeerCountersAtomic> for PeerCounters {
     }
 }
 
-impl From<&Peer> for PeerStats {
-    fn from(peer: &Peer) -> Self {
+impl PeerStats {
+    /// Builds a snapshot for one peer.
+    pub(crate) fn from_peer(peer: &Peer, include_bitfield: bool) -> Self {
         let state = peer.get_state();
         Self {
             counters: peer.stats.counters.as_ref().into(),
@@ -66,6 +92,18 @@ impl From<&Peer> for PeerStats {
             },
             client_name: match state {
                 PeerState::Live(l) => l.client_name.clone(),
+                _ => None,
+            },
+            have_pieces: match state {
+                // The bitfield is sized from total_pieces, so the count fits a
+                // u32 by construction.
+                PeerState::Live(l) if include_bitfield => {
+                    Some(u32::try_from(l.bitfield.count_ones()).unwrap_or(u32::MAX))
+                }
+                _ => None,
+            },
+            have_bitfield: match state {
+                PeerState::Live(l) if include_bitfield => Some(l.bitfield.as_raw_slice().to_vec()),
                 _ => None,
             },
         }
@@ -96,4 +134,22 @@ impl PeerStatsFilterState {
 pub struct PeerStatsFilter {
     #[serde(default)]
     pub state: PeerStatsFilterState,
+    /// Populate `have_pieces` and `have_bitfield`.
+    ///
+    /// Off by default, and nothing is computed unless it is set: the bitfield
+    /// costs a byte per eight pieces per peer to copy, and counting its bits
+    /// is a scan the existing callers have no use for.
+    #[serde(default)]
+    pub include_bitfield: bool,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::PeerStatsFilter;
+
+    /// The default filter computes nothing, so existing callers pay nothing.
+    #[test]
+    fn bitfields_are_off_by_default() {
+        assert!(!PeerStatsFilter::default().include_bitfield);
+    }
 }
--- a/crates/librqbit/src/lib.rs
+++ b/crates/librqbit/src/lib.rs
@@ -99,6 +99,7 @@ pub use stream_connect::ConnectionOptions;
 pub use torrent_state::{
     ManagedTorrent, ManagedTorrentShared, ManagedTorrentState, TorrentMetadata, TorrentStats,
     TorrentStatsState,
+    live::peer::stats::snapshot::{PeerStats, PeerStatsFilter, PeerStatsFilterState},
 };
 pub use type_aliases::FileInfos;
 
--- a/crates/librqbit/src/torrent_state/live/mod.rs
+++ b/crates/librqbit/src/torrent_state/live/mod.rs
@@ -122,6 +122,19 @@ fn make_piece_bitfield(lengths: &Lengths) -> BF {
     BF::from_boxed_slice(vec![0; lengths.piece_bitfield_bytes()].into_boxed_slice())
 }
 
+/// Builds a peer's bitfield from the bytes it sent, clearing the padding.
+///
+/// A bitfield is byte-padded, so the bits past `total_pieces` are spare. The
+/// spec says a peer zeroes them, but only the byte length is validated, so a
+/// peer can set them. Clearing them here rather than working around them at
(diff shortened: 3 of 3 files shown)


crates/librqbit/src/storage/filesystem/fs.rs
```diff
@@ -16,6 +16,25 @@ use crate::storage::{StorageFactory, TorrentStorage};
...
+///
+/// Windows refuses an open whose access conflicts with the sharing mode an
+/// existing handle was created with. Asking for write access on a file another
+/// application opened with `FILE_SHARE_READ` is exactly that conflict.
+///
+/// `ERROR_SHARING_VIOLATION` (32) and `ERROR_LOCK_VIOLATION` (33) are the two
+/// this can surface as. Nothing equivalent happens on Unix, where an open
+/// handle does not restrict other opens, so the check is compiled out.
+#[cfg(windows)]
+fn is_sharing_violation(e: &std::io::Error) -> bool {
```
ikatson: this seems to be from another PR https://github.com/ikatson/rqbit/pull/645, let's split

crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
```diff
@@ -68,10 +93,32 @@ impl From<&Peer> for PeerStats {
...
 
+/// Counts the pieces a peer claims, ignoring the bitfield's padding.
+///
+/// The trailing bits past `total_pieces` are spare. The spec says a peer zeroes
+/// them, but nothing enforces that, so they are excluded rather than trusted —
+/// the same slicing `on_bitfield` already does for its "peer has full torrent"
+/// check. `min` guards the slice: the byte length is validated on ingest, so
+/// the two should agree, but a panic here would take down a stats call.
+fn count_have_pieces(bitfield: &crate::type_aliases::BF, total_pieces: u32) -> u32 {
+    let end = (total_pieces as usize).min(bitfield.len());
```
ikatson: can we just fix the original bitfield length "at ingest" as it's called in description instead of here?

crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
```diff
@@ -68,10 +93,32 @@ impl From<&Peer> for PeerStats {
...
                 _ => None,
             },
+            have_pieces: match state {
+                PeerState::Live(l) if include_bitfield => {
+                    Some(count_have_pieces(&l.bitfield, total_pieces))
+                }
+                _ => None,
+            },
+            have_bitfield: match state {
+                PeerState::Live(l) if include_bitfield => Some(l.bitfield.as_raw_slice().to_vec()),
```
ikatson, line 106: why convert to vec and not just clone?

crates/librqbit/src/storage/filesystem/fs.rs
adamgreenwell: Split. That commit was on this branch because Flume needs both patches and `[patch.crates-io]` takes a single source — my problem rather than yours, and there's a separate integration branch carrying both now.

#644 is the bitfield work only, rebased onto main. The Windows change lives in #645 alone.

crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
adamgreenwell: Done — `on_bitfield` clears the padding now, `count_have_pieces` is gone, and `count_ones()` is correct everywhere. `from_peer` lost its `total_pieces` parameter as a result, so the diff got smaller.

One thing it turned up, worth a look: `on_have` is a second way in. It bounds the index with `get_mut`, which checks against the bitfield's *length* — byte-padded, so up to 7 over the piece count — so a peer could set a padding bit one `Have` after ingest and put the count back where it was. It's bounded by the piece count now, in the same commit. That also stops a peer claiming a piece that does not exist.

Neither was harmful before this: every other reader either slices to `total_pieces` first, as `on_bitfield`'s own "peer has full torrent" check does, or indexes by an already-validated piece id. Which is presumably why it went unnoticed — the padding is only visible to something that counts the whole bitfield.

It's a separate commit (`Clear a peer bitfield's padding on ingest`) so you can take it on its own, independently of the stats work.

crates/librqbit/src/torrent_state/live/peer/stats/snapshot.rs
adamgreenwell: `PeerStats` derives `Serialize` and `bitvec` is built without its `serde` feature, so `BitBox` has no `Serialize` impl — `l.bitfield.clone()` does not compile.

Enabling the feature would also put `bitvec` in the public API: a dependent would need a matching `bitvec` version just to name the field's type. The allocation is the same either way, so `Vec<u8>` looked like the cheaper commitment. I have put that reasoning in a comment on the field, since it is not obvious from the call site.

PR #646, opened 2026-09-02
fix `request_one` dupe call in dht

moshen:
I'm guessing that this is unintentional, in that on [line 431](https://github.com/ikatson/rqbit/blob/a499d2f243d124e144aef137afe7cb304a6e3f36/crates/dht/src/dht.rs#L429-L435) we're already issuing the request.

Found while noodling around with this library and seeing dht errors come in pairs.


ikatson:
Thanks, good catch! It does indeed seem unintentional

PR #647, opened 2026-09-02
feat(streaming): make per-stream lookahead window configurable

zond:
The per-stream streaming lookahead was hardcoded as a private 32MiB const (PER_STREAM_BUF_DEFAULT), so callers could not tune how far ahead of the reader pieces are prioritized.

Replace it with a public DEFAULT_STREAM_LOOKAHEAD_BYTES const and a FileStreamOptions { lookahead_bytes } struct whose Default preserves the 32MiB value. Add ManagedTorrent::stream_with_options and Api::api_stream_with_options; the existing stream()/api_stream() now delegate with the default, so every current caller is unchanged. The HTTP streaming endpoint gains an optional lookahead_bytes query param (absent = default, 0 = 400 Bad Request).

Behavior is identical for existing callers: the default field produces the same piece range as before, and start + N becomes start.saturating_add(N) purely as overflow hardening.

Created by asking Claude Fable 5 to fix this with a lot of opinions about code style and quality.

### DIFF
--- a/crates/librqbit/src/api.rs
+++ b/crates/librqbit/src/api.rs
@@ -16,7 +16,7 @@ use crate::{
     },
     session_stats::snapshot::SessionStatsSnapshot,
     torrent_state::{
-        FileStream, ManagedTorrentHandle,
+        FileStream, FileStreamOptions, ManagedTorrentHandle,
         peer::stats::snapshot::{PeerStatsFilter, PeerStatsSnapshot},
     },
     type_aliases::BF,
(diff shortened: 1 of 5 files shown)


ikatson:
I'm confused as to why this is the same as https://github.com/ikatson/rqbit/pull/640, but submitted by a different account. Are you copying other people's PRs and advertising as your own?

zond:
This is me not checking my agents work enough - it found the need for this feature, and when I asked it to implement it in my fork it also found https://github.com/ikatson/rqbit/pull/640, and decided https://github.com/ikatson/rqbit/pull/640 already did it the best way.

Then when I wanted to contribute it upstream I didn't check if there were already identical PRs.

Apologies, it's all on me.

PR #651, opened 2026-09-06
fix(peer): detect peer capabilities from the peer's handshake on incoming connections

zond:
On an incoming connection we built our own handshake and then read the extension bits back out of it, so every incoming peer was assumed to support BEP-10 regardless of what it actually sent. The peer's handshake was already available (it is used for the info hash check just above), it just wasn't consulted.

Also pass the peer's handshake to on_handshake(), which is what every implementation of it expects (the outgoing path already does this).

Created by my friendly Claude Opus 5


ikatson:
thanks for the fix!

PR #652, opened 2026-09-07
fix(storage): don't hide or lose is_type_id

zond:
My agent claims this is a bug, and it kind of seems to be when I look at the code - the is_type_id seems to be meant to return the bottom level storage type, in which case middlewares shouldn't return their own types.

Just close it if I'm wrong I guess.


ikatson:
this is all sort-of example code not enabled by default, so doesn't matter that it's bugged, but the fix is legit, so merging, thanks!

zond:
Did you forget the actual merge? :D

PR #660, opened 2026-09-07
fix(bencode): limit deserializer recursion depth (remote stack-overflow abort)

tweedge:
## Problem

The bencode deserializer recurses with no depth limit, so a deeply nested payload of about 20 KB overflows the stack and kills the whole process. This is a fatal abort, not a catchable panic:

```
fatal runtime error: stack overflow, aborting
```

I found this while looking at rqbit and working on a personal project. Every remotely-supplied bencode flows through this one deserializer, so all of these surfaces are affected:

- `.torrent` files and magnet metadata (BEP 3/9) via `torrent_from_bytes`
- peer extended messages (BEP 10) via `peer_binary_protocol::extended`
- DHT KRPC packets (BEP 5) via `dht::bprotocol::deserialize_message`
- tracker HTTP responses via `tracker_comms`

<details>
<summary>Reproducer (pre-fix)</summary>

```rust
// on a 2 MiB-stack thread (tokio's default)
let mut buf = vec![b'l'; 20_000];
buf.extend(std::iter::repeat_n(b'e', 20_000));
let v: Result<BencodeValue<ByteBufOwned>, _> = from_bytes(&buf);
// fatal runtime error: stack overflow, aborting
```

Also reproduces through `torrent_from_bytes` with deep nesting in an unknown top-level field, since serde's `IgnoredAny` path recurses too. Verified at depth 20k, 60k, and 200k.

</details>

## Fix

I found recursion in two places: `deserialize_seq` and `deserialize_map` (`deserialize_struct`, `deserialize_tuple[_struct]`, and `deserialize_ignored_any` all delegate to them). Guarding those two guards every container path, and it also bounds the recursive `Drop` of `BencodeValue`/`TorrentMetaV1`, which is a second overflow path an iterative parser wouldn't fix.

The change is a depth counter on `BencodeDeserializer` with a public `DEFAULT_MAX_DEPTH = 128` and a `with_max_depth()` builder, plus a new `Error::DepthLimit(usize)` variant. Serialization is untouched since parsed depth is now bounded and serialized values are constructed in-process.

## Why 128

Bencode is shallow in practice: a `.torrent` peaks around depth 6, and DHT, tracker, and extended messages reach 3 or 4. BEP 3 technically doesn't define anything about nesting, but libtorrent's `bdecode()` defaults to 100 and serde_json caps at 128, so I went with 128 and left `with_max_depth()` as an escape hatch. Worst case is tens of KB of stack, which is safe on tokio's 2 MiB workers.

## Performance

One `usize` increment and compare per container start, nothing else. Parsing a real 215 KB `.torrent` 300 times in release: 25.8 ms vs 25.4 ms on 9.0.1, which is run-to-run noise.

## Tests

Parsing at exactly the limit works, limit+1 returns `DepthLimit(128)`, `with_max_depth` is honored, and a 1 MB payload of bare `l` bytes errors cleanly instead of overflowing. The bencode suite passes 21/21, and the lib tests for librqbit-core, peer-binary-protocol, dht, and tracker-comms all pass.

### DIFF
--- a/crates/bencode/src/deserialize.rs
+++ b/crates/bencode/src/deserialize.rs
@@ -8,11 +8,26 @@ use serde::{Deserialize, Serialize, forward_to_deserialize_any};
 
 use crate::raw_value::TAG;
 
+/// The default maximum nesting depth accepted when deserializing.
+///
+/// Bencode itself (BEP 3) places no limit on nesting, so this is a
+/// receiver-side robustness limit, not a format limit. The deserializer is
+/// recursive, so without it an attacker-controlled payload of a few tens of
+/// kilobytes (e.g. a deeply nested tracker response, a malformed .torrent
+/// file, a BEP 10 extended message, or a DHT packet) overflows the stack and
+/// aborts the process. Legitimate bencode in the BitTorrent ecosystem is
+/// shallow (a .torrent peaks around depth 6; DHT and tracker messages around
+/// 3-4), so 128 leaves ~20x headroom over anything real. This matches
+/// serde_json's recursion limit; libtorrent's bdecode() defaults to 100.
+pub const DEFAULT_MAX_DEPTH: usize = 128;
+
 pub struct BencodeDeserializer<'de> {
     buf: &'de [u8],
     field_context: ErrorContext<'de>,
     field_context_did_not_fit: u8,
     parsing_key: bool,
+    depth: usize,
+    max_depth: usize,
 }
 
 impl<'de> BencodeDeserializer<'de> {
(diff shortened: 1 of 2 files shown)


crates/bencode/src/deserialize.rs
```diff
@@ -359,7 +401,10 @@ impl<'de> serde::de::Deserializer<'de> for &mut BencodeDeserializer<'de> {
         V: serde::de::Visitor<'de>,
     {
         self.parse_first_byte(b'l', Error::InvalidValue)?;
-        visitor.visit_seq(SeqAccess { de: self })
+        self.enter_container()?;
+        let r = visitor.visit_seq(SeqAccess { de: self });
+        self.depth -= 1;
```
ikatson: This probably won't do anything though? 
It's incremented then decremented right away here and below. The decrement should happen when the container is finished, not when it's created

ikatson:
Thanks for the fix!

Issue #634, opened 2026-08-19
`rust-tls` feature compiles aws-lc-rs alongside ring — SHA-1 backend can't be chosen independently of TLS


maxiarat1:
## Summary

The `sha1-ring` feature is wired to `aws-lc-rs`, not `ring`:

```toml
# crates/sha1w/Cargo.toml (v8.1.1 and still in v9.0.0)
[features]
default = ["sha1-crypto-hash"]
sha1-crypto-hash = ["crypto-hash"]
sha1-ring = ["aws-lc-rs"]
```

and `rust-tls` bundles the SHA-1 backend choice with the TLS choice:

```toml
# crates/librqbit/Cargo.toml
rust-tls = ["reqwest/rustls", "sha1w/sha1-ring", "librqbit-core/sha1-ring"]
```

## Effect on downstream builds

For a consumer using `librqbit = { default-features = false, features = ["rust-tls"] }` whose dependency graph resolves rustls with the `ring` provider, both crypto stacks get compiled:

- `aws-lc-sys` is the single most expensive compilation unit in our workspace build (~22.6 CPU-seconds, ~7% of the total), pulled in solely for SHA-1 hashing via `sha1w`;
- `ring` is compiled anyway for rustls;
- the two libraries implement the same primitives.

There is no feature combination that avoids this: `rust-tls` hard-couples `sha1w/sha1-ring`, and the alternative (`default-tls` → `sha1-crypto-hash`) lands on native-tls/OpenSSL, which is a portability regression on Linux for otherwise pure-Rust builds.

## Suggestions (any one of these would resolve it)

1. Decouple the SHA-1 backend from the TLS backend — e.g. let consumers select `rust-tls` for transport and `sha1-crypto-hash` (or a new pure-Rust `sha1` backend such as the RustCrypto `sha1` crate) independently.
2. Make `sha1-ring` actually use `ring` (`ring::digest::SHA1_FOR_LEGACY_USE_ONLY` exists for exactly this), so rustls-with-ring consumers get one crypto stack.
3. At minimum, rename the feature so the `aws-lc-rs` dependency is visible in the name — the current name suggests `ring` and makes the extra crypto stack easy to miss in an audit.

Happy to test a branch against our workspace if useful.

ikatson:
Sha1 create is extremely slow, so I removed it long ago. 

Happy to take a PR which uses ring instead of Aws - that seems like a free purely positive change with no breakage and reduced deps and compilation time (option 2)

[...]

PR #662, opened 2026-09-07
fix(sha1w): make sha1-ring use ring, decouple SHA-1 backend from TLS

tweedge:
## Problem

`sha1-ring` is wired to `aws-lc-rs`, not `ring`:

```toml
# crates/sha1w/Cargo.toml
sha1-ring = ["aws-lc-rs"]
```

```toml
# crates/librqbit/Cargo.toml
rust-tls = ["reqwest/rustls", "sha1w/sha1-ring", "librqbit-core/sha1-ring"]
```

`reqwest/rustls` on reqwest 0.13 defaults to the `aws-lc-rs` provider, so a consumer asking for rustls-with-ring compiles both crypto stacks: `aws-lc-sys` pulled in solely for SHA-1 via `sha1w`, plus whatever the TLS provider pulls in. There is no feature combination that avoids it: `rust-tls` hard-couples the hash backend, and the alternative (`default-tls` -> `sha1-crypto-hash` -> OpenSSL via `crypto-hash`) is a portability regression for otherwise pure-Rust builds.

I found this while looking at the feature graph for #634. History shows the intent was always `ring`: `01a4f68e Make rust-tls work (through ring)` introduced `sha1-ring = ["ring"]`, and `e25612b7 switch ring to aws-lc-rs` flipped the backend without renaming the feature.

Reproducer (pre-fix):

```
cargo tree -p librqbit --no-default-features -F rust-tls -e features -i aws-lc-rs
# aws-lc-rs appears via both reqwest (__rustls-aws-lc-rs) and sha1w
```

## Fix

`sha1-ring` now means `ring` again. `ring::digest::{Context, SHA1_FOR_LEGACY_USE_ONLY, SHA256}` is API-compatible with the `aws-lc-rs` usage, so `crates/sha1w/src/lib.rs` is a one-line import change.

`crates/librqbit/Cargo.toml`:

```toml
default-tls = ["reqwest/default-tls", "sha1-crypto-hash"]
rust-tls = [
    "reqwest/rustls-no-provider",
    "dep:ring",
    "dep:rustls",
    "sha1-ring",
    "librqbit-upnp/rust-tls-ring",
    "tracker_comms/rust-tls-ring",
]
sha1-crypto-hash = [
    "sha1w/sha1-crypto-hash",
    "librqbit-core/sha1-crypto-hash",
]
sha1-ring = ["sha1w/sha1-ring", "librqbit-core/sha1-ring"]
```

`rust-tls` uses `reqwest/rustls-no-provider` plus direct `ring` and `rustls` (workspace: `default-features = false, features = ["std", "tls12", "ring"]`) deps. `reqwest/rustls-no-provider` deliberately enables no crypto provider, so every reqwest client construction site installs the ring provider as the process default first (`rustls::crypto::ring::default_provider().install_default()`, ignoring the lost-race error): `Session` startup, `HttpApiClient::new`, `IpRanges::load_from_url`, and shared `build_reqwest_client()` helpers in `librqbit-upnp` and `librqbit-tracker-comms` (covering UPnP port-forward/discovery and upnp-serve NOTIFY). `default-tls` keeps the reqwest default. The new top-level `sha1-crypto-hash` / `sha1-ring` features make the hash backend selectable independently of transport; `rust-tls` defaults to `sha1-ring`, `default-tls` to `sha1-crypto-hash`, preserving existing defaults.

## Why this way

Making `sha1-ring` actually use `ring` resolves the reported double-stack build with no new API and no rename churn. Decoupling the features on top of that resolves the other half of #634: transport and hashing are now orthogonal at the feature level, instead of forcing a second change later. A pure-RustCrypto `sha1` backend was left out: `lib.rs` notes hashing is the majority of librqbit CPU and OpenSSL is 2-3x faster than Rust SHA-1, so adding a slower third backend did not seem worth it here.

Note: `-F rust-tls,sha1-crypto-hash` unifies `sha1w` with both backends and trips its `assert_cfg::exactly_one!` guard, so that combination does not compile by design (cargo features are additive). The decoupling here means TLS no longer forces a hash backend; selecting the non-default hash backend is a separate opt-in for consumers that only enable one.

Tradeoff: enabling `default-tls` and `rust-tls` together unifies reqwest with `__rustls-aws-lc-rs`, so reqwest's own default-provider fallback picks aws-lc-rs while direct-rustls users see both `ring` and `aws_lc_rs` crate features. That is inherent to cargo feature unification, same as any crate enabling both providers. Also, anyone relying on `sha1w/sha1-ring` pulling in `aws-lc-rs` will now get `ring` instead; the feature name always said `ring`, so this restores the documented meaning.

## Performance / build impact

No hot-path change: `ring::digest::Context` update/finish replaces the same calls on `aws_lc_rs::digest::Context`. Build impact is the point: with `--no-default-features -F rust-tls`, `aws-lc-rs` is gone from the `librqbit` feature tree:

```
cargo tree -p librqbit --no-default-features -F rust-tls,http-api-client -e features -i aws-lc-rs
# error: package ID specification `aws-lc-rs` did not match any packages
```

That drops the `aws-lc-sys` compile unit the reporter of #634 measured at ~22.6 CPU-seconds in their workspace. `default-tls` builds still pull `aws-lc-rs` via the reqwest default, unchanged.

## Cargo.lock

Intended changes only: `sha1-wrapper` swaps `aws-lc-rs` for `ring`, `librqbit`/`upnp`/`tracker-comms` gain `ring`+`rustls` edges, `aws-lc-rs` drops its `untrusted 0.7.1` edge (only it used that version, so the duplicate entry goes away), and `tempfile` resolves `getrandom` to 0.3.4 (in range `>=0.3, <0.5`, no version bump of tempfile itself). No other package versions change.

## Tests

- `cargo check -p librqbit-sha1-wrapper --features sha1-ring --no-default-features`: pass
- `cargo check -p librqbit --no-default-features -F rust-tls,http-api-client`: pass, no `aws-lc-rs` in tree
- `cargo check -p librqbit --no-default-features -F rust-tls,sha1-crypto-hash,http-api-client`: fails only on missing `pkg-config`/OpenSSL for `crypto-hash`, same failure exists on clean `main` in this container, not caused by this change
- `cargo test -p librqbit-sha1-wrapper --features sha1-ring --no-default-features`: pass
- `cargo test -p librqbit-core --features sha1-ring --no-default-features torrent_metainfo`: 5/5 pass, including `test_deserialize_torrent_with_info_hash`
- `cargo test -p librqbit-tracker-comms --features rust-tls-ring` and `cargo test -p librqbit-upnp --lib --features rust-tls-ring`: pass, including new `test_build_reqwest_client_does_not_panic` regression tests proving a reqwest client builds under `rustls-no-provider` (previously panicked with "No rustls crypto provider is configured")
- `cargo clippy -p librqbit --no-default-features -F rust-tls,http-api-client` (also `-D warnings`), `cargo clippy -p librqbit-tracker-comms -p librqbit-upnp -p librqbit-upnp-serve --lib --tests`, and `cargo fmt --all -- --check`: clean (the `librqbit-upnp` examples have a pre-existing `tracing_subscriber::fmt` breakage also present on clean `main`)


tweedge:
To be honest, while fixing the build failures, while it "works on mine" I'm not confident enough in my Rust knowledge and going to close this PR out for the time being, will revisit if my experience level increases to being able to defend the changes/design. Don't want to send in fragile code/slop after all. Ty for the awesome code. :)

PR #663, opened 2026-09-09
Fix incoming connection listener starving its own drain

semikolon:
`task_tcp_listener` matches `Some(Ok(..)) = futs.next()`. A failed check is the common case (unknown infohash, paused torrent, blocklist), and the refutable pattern disables that branch for the rest of the `select!` call while `accept()` keeps adding, so accepted sockets accumulate unread.

Observed on a live host: ~990 fds/min, reaching 13.6k, with Recv-Q 68 on each and 628 in CLOSE_WAIT. After the fix: 17.

Matches on `Some(res)` with an explicit `Err` arm, as the persistence loop in the same file already does, plus `biased` so pending checks drain before more are accepted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2794,7 +2794,7 @@ dependencies = [
 
 [[package]]
 name = "librqbit"
-version = "8.1.0"
+version = "8.1.1"
 dependencies = [
  "anyhow",
  "arc-swap",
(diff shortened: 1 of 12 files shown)


ikatson:
There's some old commits here, please remove 

Also excessive comments, no need for any

ikatson:
Fixed in #674

PR #665, opened 2026-09-12
fix(session): don't panic in task_listener when all select! branches are disabled (SIGABRT after boot on large libraries)

tweedge:
## Problem

`Session::task_listener` (both TCP and uTP listeners) runs a `tokio::select!` over two branches: accepting new connections (guarded by `futs.len() < max_pending_incoming_handshake_checks`) and draining a `FuturesUnordered` of pending handshake checks. The drain branch pattern-matched the inner `Ok`:

```rust
Some(Ok((live, checked))) = futs.next(), if !futs.is_empty() => { ... }
```

A failed handshake check resolves to `Err(..)`, which fails that pattern and disables the branch. With the queue at the cap, both branches are then disabled and tokio panics:

```
thread 'main' panicked at crates/librqbit/src/session.rs:980:
all branches are disabled and there is no else branch
```

This is a fatal SIGABRT under `panic = "abort"` (the release profile), taking the whole daemon down.

The window is real in production: handshake checks fail while torrents are still doing their initial hash check (`live_wait_initializing`'s 5s timeout at session.rs:953), so any burst of inbound connections shortly after boot — e.g. restarting a daemon seeding a multi-TB library, where hundreds of torrents are initializing simultaneously — fills the queue to the cap (default 256) and the first failed check aborts the process. I hit this ~90 seconds after boot on a host with a 4TB library, reproducibly across restarts.

## Fix

Match on the full result, so a failed check pops the queue like a successful one:

```rust
Some(res) = futs.next(), if !futs.is_empty() => {
    let (live, checked) = match res {
        Ok(v) => v,
        Err(_) => continue, // already logged via .map_err() at push time
    };
    ...
}
```

This makes the all-branches-disabled state unreachable for any cap, and restores the cap's intended memory bound (bounded by in-flight checks — failures previously wedged slots open, which is why downstream workarounds set the cap to `usize::MAX`).

Also clamp `max_pending_incoming_handshake_checks` to at least 1: with a cap of 0 (a public `ListenerOptions` field) both branches were disabled on the very first `select!` poll, panicking deterministically even without any traffic.

## Why this shape

A generic `else` branch on the `select!` would mask the condition instead of fixing it, and never-disabling accept (shedding connections at the cap instead) changes the backpressure mechanism. Draining on any completion is also exactly what the two neighboring `select!` loops in this crate already do (the persistence-add loop and `dht_utils.rs`), so this makes `task_listener` consistent with the codebase.

## Tests

Two regression tests in `session.rs` (`StubListener` implementing the crate-internal `Accept` trait with tokio duplex streams; dropping the client half makes each check fail fast with EOF, the same path as a disconnected peer):

- `test_task_listener_survives_failing_handshake_checks_at_cap`: queue at cap=2 with failing checks — panics against the pre-fix code with the exact production message, passes with the fix
- `test_task_listener_accepts_with_zero_cap`: cap=0 — panicked on the first `select!` evaluation pre-fix

Both pass 10/10 consecutive runs. `cargo test -p librqbit` is green (34 passed, including the existing TCP and uTP e2e download tests), and `cargo fmt --check` / `cargo clippy --all-targets` are clean. Also ran the `simulate_traffic` example (10 sessions, mixed TCP/uTP) through two full boots — zero panics with failed handshake checks being drained continuously.

### DIFF
--- a/crates/librqbit/src/session.rs
+++ b/crates/librqbit/src/session.rs
@@ -972,6 +972,9 @@ impl Session {
         l: A,
         max_pending_incoming_handshake_checks: usize,
     ) -> anyhow::Result<()> {
+        // Cap of 0 would make both select! branches below permanently disabled,
+        // which panics ("all branches are disabled").
+        let max_pending_incoming_handshake_checks = max_pending_incoming_handshake_checks.max(1);
         let mut futs = FuturesUnordered::new();
         let session = Arc::downgrade(&self);
         drop(self);
@@ -1002,7 +1005,16 @@ impl Session {
                         }
                     }
                 },
-                Some(Ok((live, checked))) = futs.next(), if !futs.is_empty() => {
+                Some(res) = futs.next(), if !futs.is_empty() => {
+                    // Match on the full result. Matching only Ok here would
+                    // disable the branch on Err: with the queue at the cap
+                    // that leaves both select! branches disabled, which
+                    // panics. The queue itself pops on any completion.
+                    let (live, checked) = match res {
+                        Ok(v) => v,
+                        // Already logged in the map_err() above.
+                        Err(_) => continue,
+                    };
                     let (addr, kind) = (checked.addr, checked.kind);
                     if let Err(e) = live.add_incoming_peer(checked) {
                         warn!(?addr, ?kind, "error handing over incoming connection: {e:#}");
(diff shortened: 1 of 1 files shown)


ikatson:
Fixed in #674

PR #667, opened 2026-09-17
fix(torrent): bound the peers map and don't accept incoming peers for finished torrents with upload disabled (unbounded memory growth)

### LINKED ISSUE #525: Memory leak / FD exhaustion

tweedge:
## Problem

`PeerStates::states` grows without a bound. Entries are created for every incoming connection and every address learned from trackers/DHT/PEX, but entries of peers that died or are no longer needed were never removed — the only removal path is `drop_peer` when an *outgoing* peer's backoff is exhausted. Incoming peers are never retried and usually have ephemeral source ports, so on hot torrents the map accumulates dead entries forever. On the machine from #525 this was ~60k tracked peers (`seen=1.17M`), hundreds of MB of RSS.

Second part of #525: a fully downloaded torrent in a session with upload disabled still accepts incoming connections, runs the full handshake, and keeps a socket plus a map entry alive for nothing. The outgoing path already skips such peers in `task_peer_adder()`; the incoming path had no such check.

## Fix

Two changes, both in `torrent_state/live`:

1. **Bound the peers map.** `PeerStates::prune_useless_peers()` evicts entries that can never be used again, checked whenever the map reaches `MAX_TRACKED_PEERS` (1024) from either insertion path, making the cleanup amortized O(1) per addition. Eviction order: `NotNeeded` peers with no outgoing address first (never retried, ephemeral ports — there is no way to ever talk to them again), then `Dead` incoming-only peers, then revivable `NotNeeded` outgoing peers as a last resort (re-learnable from trackers/DHT). Live, connecting, queued and dead outgoing peers are never touched, so reconnect and backoff behavior is unchanged. Removals are conditional (`DashMap::remove_if`), so a peer that concurrently became alive again (reconnected, or revived by `update_only_files`) is not dropped, and the aggregate stats counters stay consistent.

2. **Gate incoming peers like outgoing ones.** `add_incoming_peer()` now returns `NotAcceptingPeers` under the same predicate `task_peer_adder()` uses (`disable_upload() && is_finished_and_no_active_streams()`), closing the socket instead of handing it over. Note this also stops serving `ut_metadata` to magnet peers for such torrents, consistent with what the feature already does for bitfields.

## Why 1024

The cap only has to hold garbage, not live capacity: live peers are already bounded by the per-torrent semaphore (default 128), and dead outgoing peers are already dropped when their backoff (24h total delay) runs out. For comparison with how other clients bound their known-peer pools:

- libtorrent (the engine behind qBittorrent/Deluge) caps the per-torrent list of known peers at `max_peerlist_size` = **3000** by default, evicting entries once the list passes 90% of the limit, and stops adding entirely at the limit.
- Transmission caps its per-torrent connectable pool at `peer_limit * 3`, and tightens it to `peer_limit * 1` once the torrent is done, pruning once a minute by inactivity and usefulness — the same "tighten the pool for finished torrents" idea as the gate in this PR.

rqbit's 1024 is within that range (8× the default live-peer budget vs Transmission's 3×), and the worst case is now bounded: measured 572 bytes per tracked entry in release mode, so a completely garbage-filled map costs ~0.6 MiB per torrent (~12 MiB for a 20-torrent session), versus the unbounded growth reported in the issue.

## Tests

- Unit tests for the eviction order, the "never touches live/queued/dead-outgoing" invariant, counter consistency after pruning, and the `add_if_not_seen` trigger (trackers/DHT/PEX insert through it too).
- Two e2e tests (`tests/e2e_incoming_flood.rs`) that flood a real session with thousands of raw-handshake TCP connections the way a hot torrent does:
  - `test_e2e_incoming_flood_bounded`: flood targets a live torrent while the session downloads it from a seeder; ~6k peers tracked during the flood, map stays ≤ ~850 entries, listener still processes connections afterwards, FD growth bounded.
  - `test_e2e_incoming_flood_finished_no_upload` (needs `--features disable-upload`): the exact scenario from the issue — a finished torrent with upload disabled tracks zero peers under flood.
  - Both fail against the pre-fix code (5613 / 3804 tracked peers respectively), pass 3/3 consecutive runs.
- `cargo test -p librqbit` green (38 passed without the feature, 39 with), `cargo fmt --check` and `cargo clippy --all-targets` clean.

Closes #525.


ikatson:
The intention is noble, the issue is legit and known for a while, but the PR is giant, so will take some time to review

tweedge:
Completely understood and no rush :sweat_smile: 

Let me know if you need anything anytime!

PR #668, opened 2026-09-18
fix(session): don't panic the incoming-connection listener

rmathew1973:
`task_listener` panics and the task dies. The process stays up and everything else about the session looks fine, so the only sign is one panic line in the log. Meanwhile the session has stopped accepting incoming peers, and because the task owns the listening socket, the port closes with it.

The `select!` has two branches and no `else`:

```rust
r = l.accept(), if futs.len() < max_pending_incoming_handshake_checks => { ... }
Some(Ok((live, checked))) = futs.next(), if !futs.is_empty() => { ... }
```

Both switch off at the same moment. The accept branch is disabled by its precondition once `futs` reaches `max_pending_incoming_handshake_checks` (256 by default). The other is disabled when the next completed check resolves to `Err`, because `Some(Ok(..))` does not match and tokio drops a branch whose pattern fails. With nothing left and no `else`, it panics:

```
thread 'tokio-rt-worker' panicked at librqbit-9.0.1/src/session.rs:980:13:
all branches are disabled and there is no else branch
```

The `Err` half is ordinary, not exceptional. `check_incoming_connection` returns errors for "didn't find a matching torrent", connecting to ourselves, handshake read timeouts and "torrent is not live". So a session that reaches 256 in-flight handshake checks dies on the next error after that.

`Session::spawn` is fire and forget, so nothing brings the listener back.

## Fix

`else => continue`. When both branches are disabled the errored future has already been consumed, so `futs` has shrunk and the next iteration re-enables accept. No busy loop, and no work is lost because `map_err` already logged the error.

## Reproduction

Shape only, no networking needed:

```rust
let mut futs = FuturesUnordered::new();
futs.push(async { Err::<(), &str>("didn't find a matching torrent") });
let accept = std::future::pending::<()>();
tokio::pin!(accept);
tokio::select! {
    _ = &mut accept, if futs.len() < 1 => {},
    Some(Ok(())) = futs.next(), if !futs.is_empty() => {},
}
// panicked: all branches are disabled and there is no else branch
```

Hit on 9.0.1 in a home media box: the box kept serving HTTP fine for hours while quietly refusing every inbound peer. `cargo check -p librqbit --no-default-features --features rust-tls` passes with the change.


ikatson:
Fixed in #674, this commit was included

PR #671, opened 2026-09-19
Move completed torrents to another folder

### LINKED ISSUE #599: Request: ability to move downloaded torrents' files to another location
The ability to move completed torrents files to another location is a popular feature amongst other torrent clients.
It allows to easily browse available content at the filesystem level while continuing to seed.

It also allows other processes to be notified via `inotify(7)`

denisnikulin:
Adds `--move-completed-to <folder>` (`RQBIT_MOVE_COMPLETED_TO`, `SessionOptions::move_completed_to`). When a torrent finishes downloading, its files move from the output folder to that folder, keeping their sub-folder, and it keeps seeding from there.

Closes #599.

- ~~Includes #621 (release file handles on pause)~~ — merged since, so this is rebased on top of it. Resuming after such a pause is still fixed here: it otherwise fails with "file is None".
- Files are renamed, or copied and removed across filesystems. An existing destination is an error unless overwrite is allowed, as when creating files.
- Pause and delete are refused while a torrent is being moved, and the web UI shows "Moving".
- The new folder is saved to session persistence (JSON and Postgres).
- Covered by new e2e tests, and tested on a server moving from a local disk to an sshfs mount.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


denisnikulin:
not sure if this feature it needed - can re-open later

PR #680, opened 2026-09-21
feat: release handles on pause and reopen completed files read-only

exbit:
## Summary

This PR changes how `FilesystemStorage` manages file handles:

1. **Handles are (re)opened lazily.** A closed handle is reopened on the next read or write, and a read-only handle is upgraded before a write.
2. **Completed files are reopened read-only.** `live` already had a `// if we have all the pieces of the file, reopen it read only` comment, but nothing did it. This PR implements it through a new storage callback, so finished data is only held open for reading while seeding.
3. **File handles are released when a torrent is paused.** This brings back `release_files()` reverted in d7bfc7b0. With lazy reopening, resuming a paused torrent no longer runs into closed handles.

### Lazy file handles (`opened_file.rs`, `fs.rs`)
- `OpenedFile` tracks whether its fd is read-only and gets `ensure_opened_read_only()`, `ensure_opened()` and `close()`.
- `pread_exact` reopens a closed handle read-only. `pwrite_all`, `pwrite_all_vectored` and `ensure_file_length` (re)open it read/write and create the parent directory if needed.
- A read-only fd is reopened before writing. This is needed for v1 torrents without padding files: a piece on a file boundary is written into two files, and the neighbouring file may already be completed (read-only). Writing through a read-only fd fails with `EBADF` on Unix and `os error 5` on Windows, which errors the whole torrent.
- Windows: open options set `FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE` explicitly (same as the std default).

### Reopen completed files read-only
- New `TorrentStorage::on_file_completed(file_id)` with a default no-op, delegated for `Box<U>`.
- `FilesystemStorage` implements it with `OpenedFile::reopen_read_only()`.
- It is called:
  - in `initializing`, after file lengths are set, for every file whose pieces are all present. A failure is logged as a warning;
  - in `live`, when `update_file_have_on_piece_completed` reports no remaining pieces for a file. A failure is logged at debug level.
- If a reopen fails, the next read reopens the file read-only on its own, and the next write reopens it read/write.

### Release file handles on pause
- `TorrentStorage::release_files()` has a default no-op, is delegated for `Box<U>`, and is implemented by `FilesystemStorage` by closing all handles.
- It is called when a live torrent is paused, when the initial check finishes with the torrent paused, and when a pause is requested during the initial check.
- `pause()` on an already paused torrent now releases handles again and returns `Ok` instead of `torrent is already paused`.


crates/librqbit/src/storage/filesystem/fs.rs
```diff
@@ -167,3 +180,29 @@ impl TorrentStorage for FilesystemStorage {
         Ok(())
     }
 }
+
+fn writable_open_options() -> OpenOptions {
```
ikatson, line 184: some of the functions here and in opened_file are duplicates

crates/librqbit/src/storage/filesystem/fs.rs
```diff
@@ -167,3 +180,29 @@ impl TorrentStorage for FilesystemStorage {
...
+
+fn apply_share_mode(options: &mut OpenOptions) {
+    #[cfg(windows)]
+    {
+        use std::os::windows::fs::OpenOptionsExt;
+
+        options.share_mode(WINDOWS_SHARE_READ_WRITE_DELETE);
+    }
+
+    #[cfg(not(windows))]
```
ikatson, line 206: unnecessary block

crates/librqbit/src/storage/filesystem/fs.rs
```diff
@@ -16,6 +16,9 @@ use crate::storage::{StorageFactory, TorrentStorage};
 
 use super::opened_file::OpenedFile;
 
+#[cfg(windows)]
+const WINDOWS_SHARE_READ_WRITE_DELETE: u32 = 0x0000_0001 | 0x0000_0002 | 0x0000_0004;
```
ikatson, line 20: duplicate and it also seems to be default in Rust anyway

crates/librqbit/src/storage/filesystem/opened_file.rs
```diff
@@ -159,6 +169,94 @@ impl OpenedFile {
...
+    pub fn close(&self) {
+        let mut g = self.file.write();
+        g.fd = None;
+        #[cfg(windows)]
+        {
+            g.tried_marking_sparse = false;
+        }
+    }
+
+    pub fn reopen_read_only(&self) -> anyhow::Result<()> {
```
ikatson: ALL of readonly stuff seems be only be relevant on windows? Can we remove all the associated overhead for non-windows? Reopening etc

crates/librqbit/src/storage/filesystem/fs.rs
```diff
@@ -59,15 +62,14 @@ impl FilesystemStorage {
 
 impl TorrentStorage for FilesystemStorage {
     fn pread_exact(&self, file_id: usize, offset: u64, buf: &mut [u8]) -> anyhow::Result<()> {
-        self.opened_files
-            .get(file_id)
-            .context("no such file")?
-            .lock_read()?
-            .pread_exact(offset, buf)
+        let file = self.opened_files.get(file_id).context("no such file")?;
+        file.ensure_opened_read_only()?;
```
ikatson, line 66: double locking here and below

crates/librqbit/src/storage/filesystem/opened_file.rs
```diff
@@ -111,6 +114,12 @@ struct OpenedFileLocked {
...
     path: PathBuf,
     fd: Option<File>,
+    // Whether the currently opened `fd` was opened in read-only mode.
+    // A read-only fd cannot be written to: doing so fails with EBADF on unix
+    // and ERROR_ACCESS_DENIED (os error 5) on Windows. We must reopen it for
+    // writing before any write (e.g. a piece that spans into an already
+    // completed, reopened-read-only neighbouring file).
+    read_only: bool,
     #[cfg(windows)]
     tried_marking_sparse: bool,
```
ikatson, line 124: both "read_only" and "tried_marking_sparse" make sense only on windows

crates/librqbit/src/storage/filesystem/opened_file.rs
```diff
@@ -159,6 +169,94 @@ impl OpenedFile {
         })
     }
 
+    pub fn ensure_opened_read_only(&self) -> anyhow::Result<()> {
```
ikatson, line 172: lots of duplicate code here and below

crates/librqbit/src/storage/filesystem/opened_file.rs
```diff
@@ -159,6 +169,94 @@ impl OpenedFile {
...
+
+        let mut g = self.file.write();
+        if g.fd.is_some() && !g.read_only {
+            return Ok(());
+        }
+        if g.path.as_os_str().is_empty() {
+            anyhow::bail!(Error::FsFileIsNone);
+        }
+        let parent = g.path.parent().context("bug: no parent")?;
+        std::fs::create_dir_all(parent)
```
ikatson, line 216: unnecessary as this was done in storage init

crates/librqbit/src/torrent_state/mod.rs
```diff
@@ -490,7 +510,12 @@ impl ManagedTorrent {
                 Ok(())
             }
             ManagedTorrentState::Paused(_) => {
-                bail!("torrent is already paused");
+                let paused = g.state.take().assert_paused();
```
ikatson, line 513: why this change?

ikatson (changes requested):
Very dirty PR with lots of unnecessary and duplicate stuff.
My biggest problem was the one why I removed this in the first place - Windows-specific code pollutes everything too much. This CAN be done in a clean way, however neither my original implementation, nor this one is clean.
On Mac/Linux, NONE of this is needed. Even on Windows, it probably only is a problem for executables?

